PR #3, opened 2012-07-13
Garbage collection

appaquet:


mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -26,39 +29,38 @@ class MysqlStorage(name: String, host: String, database: String, username: Strin
 
   def getStorageTransaction = new MysqlTransaction(this)
 
+  def returnTransaction(trx: MysqlTransaction) {
```
coderunner: Not sure I understand the link between the name of the function and the code in it

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -208,6 +211,132 @@ class MysqlStorage(name: String, host: String, database: String, username: Strin
     }
   }
 
+  object GarbageCollector extends Instrumented {
+    private val metricCollect = metrics.timer("gc-collect")
+    private val metricCollected = metrics.counter("gc-collected")
```
coderunner: gc-collected-records ?

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -208,6 +211,132 @@ class MysqlStorage(name: String, host: String, database: String, username: Strin
...
+
+    val VERSIONS_BATCH_COUNT = 100
+    val MIN_COLLECTION = 10
+    val GC_DELAY = 1000
+
+    val killed = new AtomicBoolean(false)
+    val active = new AtomicBoolean(true)
+    var lastTokens = mutable.Map[Table, Long]()
+    var versionsCache = mutable.Map[Table, mutable.Queue[VersionRecord]]()
+    var allTables: Seq[Table] = model.allHierarchyTables
```
coderunner: val?

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -208,6 +211,132 @@ class MysqlStorage(name: String, host: String, database: String, username: Strin
...
+          info("Garbage collector shouldn't be running anymore. Stopping!")
+          throw new InterruptedException()
+        }
+
+        if (active.get()) {
+          val curCount = mutationsCount.get()
+          val mutationsDiff = curCount - lastCollectMutationsCount
+          lastCollectMutationsCount = curCount
+
+          if (mutationsDiff > 0 || lastCollected > 0 || nbEmptyCollect < 5) {
```
coderunner: Why a condition on emptyCollection?

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -208,6 +211,132 @@ class MysqlStorage(name: String, host: String, database: String, username: Strin
...
+    }, GC_DELAY, GC_DELAY, TimeUnit.MILLISECONDS)
+
+    def start() {
+      this.active.set(true)
+    }
+
+    def stop() {
+      this.active.set(false)
+    }
+
```
coderunner, line 255: Do we really need stop and kill method? Seems only kill is called from outside

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -208,6 +211,132 @@ class MysqlStorage(name: String, host: String, database: String, username: Strin
...
+    var lastTokens = mutable.Map[Table, Long]()
+    var versionsCache = mutable.Map[Table, mutable.Queue[VersionRecord]]()
+    var allTables: Seq[Table] = model.allHierarchyTables
+
+    var lastCollectMutationsCount = 0
+    var lastCollected = 0 // always force first collection
+    var nbEmptyCollect = 0 // succeeding empty garbage collection
+
+    val scheduledExecutor = new ScheduledThreadPoolExecutor(1)
+    scheduledExecutor.scheduleWithFixedDelay(new Runnable {
```
coderunner: This returns a scheduledFuture that can be canceled is the GC is killed or stop otherwise the run method will still be called periodically

coderunner: This can also be used to schedule only when the collector is started so that no check to active.get is necessary

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -208,6 +211,132 @@ class MysqlStorage(name: String, host: String, database: String, username: Strin
...
+    }
+
+    def kill() {
+      this.stop()
+      killed.set(true)
+    }
+
+    /**
+     * Garbage collect at least specified versions (if any can be collected)
+     * Not thread safe! Call from 1 thread at the time!
```
coderunner, line 263: Using an Actor for the collection would solve this problem.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -208,6 +211,132 @@ class MysqlStorage(name: String, host: String, database: String, username: Strin
...
+
+              // didn't fetch anything, no more versions after this token. rewind to beginning
+              if (fetched.size == 0)
+                lastToken = 0
+            }
+
+            var collectedVersions = 0
+            while (versions.size > 0 && collectedVersions < collectPerTable) {
+              val version = versions.dequeue()
+              val toDelete = (version.generations - table.maxVersions)
```
coderunner: toDeleteCount would be more clear

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
appaquet, line 255: We will need them when we will have the synchronisation. We don't want to garbage collect while we are sending records to a remote node.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
appaquet: I don't know how to call it though... Basically, it contains code that finishes the transaction after it has been commited. In that case, increments counters... returnTransaction => closeStorageTransaction ?

coderunner:
+1

PR #4, opened 2012-07-24
Mry third level fix

appaquet:


mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
```diff
@@ -122,8 +117,8 @@ class MysqlTransaction(storage: MysqlStorage) extends StorageTransaction with In
         %4$s
         AND o.`ts` = (  SELECT MAX(ts)
                         FROM `%2$s` AS i
-                        WHERE i.`ts` <= ? AND i.`tk` = ? AND %5$s %6$s)
-              """.format(projKeys, fullTableName, outerWhereKeys, outerWhereGens, innerWhereKeys, innerWhereGens)
+                        WHERE i.`ts` <= ? AND i.`tk` = ? AND %5$s)
```
coderunner, line 140: Not sure how we can make this simpler to understand...maybe an example SQL that this generates with an explanation?! I must admit I would not be able to easily maintain this code.

coderunner:
Super tested! Good! :) +1

alexbergeron:
+1

PR #7, opened 2012-07-26
Added support for Double and Bool types

alexbergeron:


mry-core/src/main/scala/com/wajam/mry/execution/Implicits.scala
```diff
@@ -15,6 +15,9 @@ object Implicits {
       case rVal: IntValue =>
         rVal.intValue
 
+      case rVal: BoolValue =>
+        rVal.boolValue
+
```
appaquet, line 20: Missing case for DoubleValue?

coderunner:
+1

PR #8, opened 2012-08-01
Don't use generation number to delete children hierachy

appaquet:
Don't use generation number to delete children hierarchy, generate tombstone records instead so that percolation on these tables still work. Added test to make sure percolation on children records works when deleting parent records. 

ALTER TABLE `emails` DROP COLUMN `g1` ;
ALTER TABLE `followees` DROP COLUMN `g1` ;
ALTER TABLE `followees_followers` DROP COLUMN `g2` , DROP COLUMN `g1` ;
ALTER TABLE `installs` DROP COLUMN `g1` ;
ALTER TABLE `networks` DROP COLUMN `g1` ;
ALTER TABLE `networks_followees` DROP COLUMN `g2` , DROP COLUMN `g1` ;
ALTER TABLE `networks_users` DROP COLUMN `g2` , DROP COLUMN `g1` ;
ALTER TABLE `users` DROP COLUMN `g1` ;
ALTER TABLE `users_networks` DROP COLUMN `g2` , DROP COLUMN `g1` ;
ALTER TABLE `users_networks_followees` DROP COLUMN `g3` , DROP COLUMN `g2` , DROP COLUMN `g1` ;


coderunner:
+1

PR #9, opened 2012-08-28
SCN timestamps integration

jgagnon1:


alexbergeron:
+1

mry-core/src/main/scala/com/wajam/mry/Database.scala
```diff
@@ -17,7 +18,16 @@ class Database(var serviceName: String = "database") extends Service(serviceName
   def analyseTransaction(transaction: Transaction): ExecutionContext = {
     val context = new ExecutionContext(storages)
     context.dryMode = true
-    context.timestamp = Timestamp.now
+
+    scn.getNextTimestamp(serviceName, (timestamps: List[Timestamp], optException) => {
```
coderunner: Let's talk about this tomorrow, but I think we are missing a piece. This call should be on the client interface to Scn not the service itself. This client interface should allow us to cache sequence number and avoid a round trip on every transactions to a possible remote Scn node.

mry-core/src/main/scala/com/wajam/mry/Database.scala
```diff
@@ -17,7 +18,16 @@ class Database(var serviceName: String = "database") extends Service(serviceName
...
+      if (optException.isDefined) {
+        error(optException.get.toString, 500)
```
coderunner: Let's review this code together.

coderunner:
+1

mry-core/src/main/scala/com/wajam/mry/Database.scala
appaquet: What was your conversion on this? I think it's SCN/NRV role to cache and avoid round trips by batching using the same method Google did theirs?

coderunner: I think the simplest solution for now is to have the SCN client manage this and decouple getNextTimestamp to the NRV exchange. The client would do the batching and then call NRV once it reaches the batchsize or a timeout. We could then generalize the mechanism and apply it to NRV later if we need.

appaquet: It's the simplest, but down the line it should be in NRV and I have an idea on how to do it. Lemme know when you have time, we'll chat about it ;)

coderunner:
You will have to defend the [Timestamp] -> [_] before +1 :D

coderunner:
+1

PR #13, opened 2012-09-26
Do not trace MRY db execution time.

mlaflamm:
The information was recorded after sending the response. The information is already available by looking at the span start/end time.


alexbergeron:
+1

coderunner:
+1

PR #15, opened 2012-10-11
Fix reply on SCN error

coderunner:


mry-core/src/main/scala/com/wajam/mry/Database.scala
```diff
@@ -80,11 +83,11 @@ class Database(var serviceName: String = "database", val scn: ScnClient) extends
...
-    this.metricExecuteLocal.time {
-
-      scn.fetchTimestamps(serviceName, (timestamps: Seq[Timestamp], optException) => {
+    val timerContext = this.metricExecuteLocal.timerContext()
+    scn.fetchTimestamps(serviceName, (timestamps: Seq[Timestamp], optException) => {
+      try {
         if (optException.isDefined) {
-          error(optException.get.toString, 500)
+          info("Exception while fetching timestamps from SCN.", optException.get)
           throw optException.get
```
alexbergeron, line 91: I'm not a big fan of if (isDefined) with option.get.

alexbergeron:
Everything looks allright. +1

PR #16, opened 2012-10-22
Split read write exec and avoid unecessary calls to SCN for read-only execution

coderunner:


appaquet:
:+1:

For the record, we could probably need to refactor to put the last local timestamp logic to the scn client. Let's keep it for future iteration!

PR #19, opened 2012-11-05
Temporary fix to override timeou for MRY Database service

coderunner:


alexbergeron:
Had a question about the useage of def, but that question was answered with you latest Edge pull request. :smile:

:+1:

PR #22, opened 2012-11-12
use USE INDEX to force mysql to use the proper index when calculating the max timestamp.

coderunner:


coderunner:
@appaquet can you double check I did not forget any query?

alexbergeron:
Looks good, but is it normal that we always use the revkey and never the PRIMARY key?

appaquet:
@coderunner: I would put USE INDEX(pk) for everything that is not using revkey so that we don't fuck up mysql deciding to change query plan when the table is too big. What do you think?
@alex-wajam: The external queries for timeline are using the PK. Everything that is not with notion of sequential in time aren't using the PK. Since the PK is using an always increasing timestamp as its first column, the innodb b-tree is almost just append only, which should have an impact on performance

alexbergeron:
Thanks, that pretty much answered all on my questions :smile:

coderunner:
@appaquet I could add the USE INDEX everywhere, but I doubt that the primary key index will cause trouble. I think the main issue in our case is that it had two indices it could use but not enough stats to know the best one.

So personally, I prefer leaving the sql clean if we can.

appaquet:
Go ahead then! :+1:

PR #27, opened 2012-11-26
remove auto increment column to avoid (im)possible replication problems

coderunner:
A value should always be set so auto increment is never triggered.


alexbergeron:
+1. Since we use SCN, we should never use auto-incremented values anyway.

PR #29, opened 2012-11-28
Make GC collection replication safe

appaquet:
Downside: only collects one version at the time, but at least queries can be replicated safely

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
@@ -251,7 +251,7 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
     /*
      * Generated SQL looks like:
      *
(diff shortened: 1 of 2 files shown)


mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
```diff
@@ -301,18 +301,16 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
...
     val sql = """
                 DELETE FROM `%1$s`
                 WHERE tk = ?
                 AND %2$s
-                AND ts < %3$d
-                ORDER BY ts ASC
-                LIMIT %4$d;
-              """.format(fullTableName, whereKeys, maxTimestamp.value, count)
+                AND ts = %3$d
+                ORDER BY ts ASC;
```
coderunner: don't think order by is needed now :)

appaquet: Oops.

coderunner:
+1

PR #34, opened 2012-12-03
Timeline feeder batch integrity

mlaflamm:


mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableTimelineFeeder.scala
```diff
@@ -91,9 +114,9 @@ class TableTimelineFeeder(storage: MysqlStorage, table: Table, val batchSize: In
   }
 
   def ack(data: Map[String, Any]) {
-    currentTimestamps = currentTimestamps filterNot (_ != getTimestampKey(data))
+    currentTimestamps = currentTimestamps filter (_ != getTimestampKey(data))
```
coderunner, line 113: This kind of change is always scary...

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableTimelineFeeder.scala
```diff
@@ -4,78 +4,101 @@ import com.wajam.nrv.Logging
...
-    debug("Getting new batch for table {}", table.depthName("_"))
-
-    // TODO: should make sure we don't call too often
-
-    // get starting timestamp
-    val timestampCursor: Timestamp = lastElement match {
-      case Some((ts, keys)) => ts
-      case None => Timestamp(context.data.get("from_timestamp").getOrElse("0").toLong)
-    }
+    if (lastEmptyTimestamp + emptyThrottlePeriod < currentTime) {
```
appaquet: Isn't this the job of Spnl to throttle the feeder if nothing was found ?? We kind of duplicate what SPNL is supposed to do here

mlaflamm: True, as we go back to one task per feeder feeder. Will remove ASAP.

appaquet:
:+1:

PR #39, opened 2012-12-04
Moar precise metrics for MRY mysql storage

appaquet:


mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -61,6 +63,8 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
           this.dropTable(table)
       }
     }
+
+    tablesMutationsCount = allTables.map(table => (table, new AtomicInteger(0))).toMap
```
mlaflamm, line 75: Not thread safe but this is not really new, the entire method is not thread safe. I suggest to add a doc comment to syncModel() saying the method is not thread safe and MUST be invoked before start(). If you are in the mood, you could validate that the storage is not started...

mlaflamm:
+1

PR #41, opened 2012-12-05
Adding the encoding column in mysql storage

appaquet:
For future support of different encoding of the data saved to mysql


coderunner:
+1

PR #42, opened 2012-12-08
fix connection leak

coderunner:


appaquet:
+1, never though setAutoCommit could throw!

coderunner:
In the case I've seen, it triggered opening a new connection which failed. Not sure why yet, but this will avoid connection leak.

PR #43, opened 2012-12-11
Paging/limit support in MRY

appaquet:


mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
```diff
@@ -11,18 +11,14 @@ trait Value extends Object with OperationSource {
 
   def equalsValue(that: Value): Boolean = this == that
 
-  def isNull = this.equalsValue(NullValue.NULL_VALUE)
+  def isNull = this.equalsValue(NullValue)
 }
 
-class NullValue() extends Value with Serializable {
+class NullValue extends Value with Serializable {
```
alexbergeron, line 17: Stupid question, but could you avoid implementing a class and a object with only a NullValue object that extends Value with Serializable

alexbergeron:
+1

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
appaquet, line 17: I could just implement an object instead... Didn't see that  ;)

appaquet, line 17: Finally, would be shotgun surgery and would change the PR all over. Good for a future PR

PR #45, opened 2012-12-12
Fixing GC

appaquet:


mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -16,7 +16,7 @@ import java.util.concurrent.{TimeUnit, ScheduledThreadPoolExecutor}
  */
 class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean = true)
   extends Storage(config.name) with Logging with Value with Instrumented {
-  var model: Model = new Model
+  var model: Model = _
```
coderunner, line 19: why this change?

appaquet, line 19: Fixes a bug. The start function checks if model ==  null and throws if the model is null because it can't start without a model.

coderunner:
+1

PR #47, opened 2012-12-13
GC optimization

appaquet:
Use range to query the table instead of doing a full table query to get records to GCs


alexbergeron:
+1

coderunner:
+1

PR #48, opened 2012-12-19
MRY refactor + various fixes

appaquet:
-New table structure using 2 tables for each mry table, one for index, one for data (blob)
-Benchmark tool for MRY MySQL
-Various fixes


mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
```diff
@@ -236,51 +256,53 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
...
                   selectMode: TimelineSelectMode = TimelineSelectMode.FromTimestamp): Seq[MutationRecord] = {
-    val projKeys = (for (i <- 1 to table.depth) yield "c.k%1$d".format(i)).mkString(",")
-    val whereKeys = (for (i <- 1 to table.depth) yield "i.k%1$d = c.k%1$d".format(i)).mkString(" AND ")
-    val joinKeys = (for (i <- 1 to table.depth) yield "c.k%1$d = l.k%1$d".format(i)).mkString(" AND ")
+    val projKeys = (for (i <- 1 to table.depth) yield "ai.k%1$d".format(i)).mkString(",")
+    val whereKeys1 = (for (i <- 1 to table.depth) yield "ai.k%1$d = ad.k%1$d".format(i)).mkString(" AND ")
+    val whereKeys2 = (for (i <- 1 to table.depth) yield "ai.k%1$d = bi.k%1$d".format(i)).mkString(" AND ")
+    val whereKeys3 = (for (i <- 1 to table.depth) yield "ci.k%1$d = ai.k%1$d".format(i)).mkString(" AND ")
+    val whereKeys4 = (for (i <- 1 to table.depth) yield "bi.k%1$d = bd.k%1$d".format(i)).mkString(" AND ")
     val fullTableName = table.depthName("_")
```
coderunner, line 264: would it make sens to compute all of these once only and store it in the Table?

coderunner, line 264: We should also turn on PreparedStatement caching since the queries will always be the same. I'll take care of that.

appaquet, line 264: Both would make sense and I was waiting for more stability before going into that kind of optimization

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
```diff
@@ -359,27 +381,32 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
...
-      }
+    /*
+     * Generated SQL looks like:
+     *
+     *     DELETE FROM `table2_table2_1_table2_1_1_data`
+     *     WHERE k1 = ? AND k2 = ? AND k3 = ?
+     *     AND ts IN (3242343243243, 34243423434);
+     */
+    val dataSql = """
+                DELETE FROM `%1$s_data`
```
coderunner, line 402: Maybe should refactor both delete statement for index and data in a function to which we pass the suffix to make sure both are always updated in sync. Maybe store result in Table.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
```diff
@@ -93,64 +93,85 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
...
+            *       GROUP BY tk, k1, k2
+            *     ON DUPLICATE KEY UPDATE d=VALUES(d)
+            */
+            val childIndexSql = """
+                 INSERT INTO `%1$s_index`
+                  SELECT ? AS ts, tk, %2$s
+                  FROM `%1$s_index`
+                  WHERE tk = ? AND ts <= ? AND %3$s
+                  GROUP BY tk, %2$s
+                 ON DUPLICATE KEY UPDATE ts=VALUES(ts)
```
coderunner, line 170: Why do we update the timestamp in child tables? Also comment and code not in sync.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -112,11 +110,12 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
...
+      if (update) {
         results.statement.executeUpdate()
-      else
+        results.close()
+      } else {
         results.resultset = results.statement.executeQuery()
-
+      }
 
       results
```
coderunner, line 120: Not sure I like this change. Maybe the public API should expose two methods and use this method with a bool as a parameter internally to avoid problems.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -139,7 +138,10 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
 
       var tables = List[String]()
       while (rs.next()) {
-        tables ::= rs.getString(1)
+        val name = rs.getString(1)
+        if (name.contains("_index")) {
+          tables ::= name.stripSuffix("_index")
+        }
```
coderunner, line 144: Shouldn't we strip _data suffix as well to be consistent?

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
coderunner, line 170: Ok get it.

coderunner:
+1

PR #50, opened 2012-12-25
Prevent full table scan on garbage collection

appaquet:


coderunner:
Can not be automerge, sinon +1

PR #51, opened 2013-01-08
Fetch MRY timestamp from new ConsistencyMasterSlave (READ DISCLAIMER)

mlaflamm:
The Timestamp trait has been moved from Scn to Nrv. This change requires a simultaneous deployment of all servers (SCN + EDGE)  because the SCN messages are not backward compatible. Deployment will be done during a period of low traffic. Merge to master will only be done once that deployment is planned.

### DIFF
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlBase.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlBase.scala
@@ -3,8 +3,8 @@ package com.wajam.mry.storage.mysql
 import com.wajam.mry.execution._
 import com.wajam.mry.storage.Storage
 import org.scalatest.{BeforeAndAfterEach, FunSuite}
-import com.wajam.scn.Timestamp
 import com.wajam.scn.storage.ScnTimestamp
+import com.wajam.nrv.utils.timestamp.Timestamp
 
 
 abstract class TestMysqlBase extends FunSuite with BeforeAndAfterEach {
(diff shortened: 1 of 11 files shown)


mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlBase.scala
```diff
@@ -56,7 +56,9 @@ abstract class TestMysqlBase extends FunSuite with BeforeAndAfterEach {
   }
 
   def createTimestamp(time: Long) = new Timestamp {
-    def value = time
+    val value = time
   }
 
+  def createNowTimestamp() = createTimestamp(System.currentTimeMillis() * 10000)
```
alexbergeron, line 62: So that was the hack you were talking about :wink: 
Maybe you could leave a todo to remember to clean this up

mry-core/src/main/scala/com/wajam/mry/Database.scala
```diff
@@ -100,62 +95,18 @@ class Database(var serviceName: String = "database", val scn: ScnClient)
...
-      val savedTimestamp = lastWriteTimestamp.get()
-      updateSuccessful = savedTimestamp match {
-        case Some(prevTimestamp) => if (timestamp > prevTimestamp) {
-          lastWriteTimestamp.compareAndSet(savedTimestamp, Some(timestamp))
-        } else {
-          true //the timestamp is less than the last timestamp so our value is obsolete
-        }
-        case None => lastWriteTimestamp.compareAndSet(savedTimestamp, Some(timestamp))
-      }
-    } while (!updateSuccessful)
```
alexbergeron, line 154: Was wondering why there was so much deleted code. Then I saw that do .. while loop :wink:

alexbergeron:
:+1:

mry-core/src/main/scala/com/wajam/mry/Database.scala
mlaflamm, line 154: Not really, just moved "as is" somewhere else...

alexbergeron, line 154: Yeah, saw it one pull request later...

PR #53, opened 2013-01-15
Better handling of protofiles

oliviertoupin:
It replicated what I did in Nrv. It makes compiling protofile less confusing. Also, if the file exists already and haven't change, it won't attempt to generate it, so no need to have protoc installed.

### DIFF
--- a/project/Project.scala
+++ b/project/Project.scala
@@ -47,6 +47,12 @@ object MryBuild extends Build {
     .settings(StartScriptPlugin.startScriptForClassesSettings: _*)
     .aggregate(core)
 
+  import sbtprotobuf.{ProtobufPlugin => PB}
+
+  val protobufSettings = PB.protobufSettings ++ Seq(
+    javaSource in PB.protobufConfig <<= (sourceDirectory in Compile)(_ / "java")
+  )
+
   lazy val core = Project("mry-core", file("mry-core"))
     .configs(IntegrationTest)
     .settings(defaultSettings: _*)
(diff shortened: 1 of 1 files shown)


project/Project.scala
```diff
@@ -56,15 +62,6 @@ object MryBuild extends Build {
...
-  import sbtprotobuf.{ProtobufPlugin => PB}
-
-  lazy val proto = Project(
-    id = "mry-proto",
-    base = file("mry-core"),
-    settings = defaultSettings ++ PB.protobufSettings ++ Seq(
-      javaSource in PB.protobufConfig <<= (sourceDirectory in Compile)(_ / "java")
-    )
-  ) configs (IntegrationTest)
+    .settings(protobufSettings: _*)
```
appaquet, line 65: Weren't we saying we wanted that on the "mry" root project so that we don't need to specify the project? Unless your fix is doing that?

oliviertoupin, line 65: You are right... :-(

But we can't put directly into mry, the path would be wrong... so we can't (unless specifing the path manually)...

So this fix doesn't fix what we talk about yesterday.

However, this:

-Is much more compact
-Associate protobuf with it's rightful project: mry-core

Up to you to accept it, or keep the old way!

appaquet, line 65: Ok then

appaquet:
+1

oliviertoupin:
I realize than protoc is running on a newly cloned repo, so we can't use that fix, on machines without protoc, I'll drop the change...

I might add a comment like in NRV instead

-2

PR #59, opened 2013-01-18
Share metrics between MysqlTransaction instances

mlaflamm:


coderunner:
+1 but is there any reason why the model is not passed in the constructor. It seems that it would make the code simpler

PR #64, opened 2013-02-08
GC truncate all extra versions before going to the next token

mlaflamm:
Also added two new metrics: 
- A gauge with the token curently GCed
- A meter with the number of extra versions loaded when a token in not collected in one GC run.

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
@@ -11,6 +11,7 @@ import java.util.concurrent.atomic.AtomicInteger
 import collection.mutable
 import java.util.concurrent.{TimeUnit, ScheduledThreadPoolExecutor}
 import com.wajam.nrv.service.TokenRange
+import com.yammer.metrics.core.Gauge
 
 /**
  * MySQL backed storage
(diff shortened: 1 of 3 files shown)


mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -317,12 +318,19 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
...
       "gc-collected-record", table.uniqueName, "records", TimeUnit.SECONDS))
+    lazy private val extraVersionsLoadedMeter = new Meter(metrics.metricsRegistry.newMeter(GarbageCollector.getClass,
+      "gc-extra-record-loaded", table.uniqueName, "extra-record-loaded", TimeUnit.SECONDS))
+    private val collectedTokenGauge = metrics.metricsRegistry.newGauge(GarbageCollector.getClass,
+      "gc-collected-token", table.uniqueName, new Gauge[Long] {
+        def value = tableNextToken
+      })
 
     private var tableNextRange: TokenRange = tokenRanges.head
+    @volatile
```
appaquet, line 333: Why? It should be called by the same timer thread everytime?

mlaflamm, line 333: This field is also read in a gauge which is called from the thread pushing metrics to graphite. May effectively be overkill since the value is only read and is to get a rough idea of where the GC is working.

coderunner:
+1

PR #65, opened 2013-02-12
Timestamp support both type + Helper

oliviertoupin:


mry-core/src/main/scala/com/wajam/mry/Database.scala
```diff
@@ -107,7 +108,7 @@ class Database(var serviceName: String = "database")
 
   private def execute(req: InMessage) {
     var values: Seq[Value] = null
-    val timestamp: Timestamp = req.metadata("timestamp").asInstanceOf[Timestamp]
+    val timestamp: Timestamp = req.metadata.getValueOldOrFromString("timestamp", (v: Seq[String]) => {Timestamp(v(0))}).asInstanceOf[Timestamp]
```
alexbergeron: getOldValueOrFromString?

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableContinuousFeeder.scala
```diff
@@ -23,12 +25,13 @@ class TableContinuousFeeder(name: String ,storage: MysqlStorage, table: Table, t
...
     val data = context.data
-    lastRecord = if (data.contains(Keys) && data.contains(Token) && data.contains(Timestamp))
+    val bool: Boolean = data.containsEither(Timestamp)
+    lastRecord = if (data.contains(Keys) && data.contains(Token) && bool)
     {
       try {
         val record = new Record()
         record.token = data(Token).toString.toLong
-        record.timestamp = com.wajam.nrv.utils.timestamp.Timestamp(data(Timestamp).toString.toLong)
+        record.timestamp = NrvTimestamp(data.getValueOldOrFromString(Timestamp, (v: Seq[String]) => {NrvTimestamp(v(0))}).asInstanceOf[NrvTimestamp])
```
mlaflamm: I don't think the conversion is needed since Record is not a Message and never put in a Message

mry-core/src/main/scala/com/wajam/mry/Database.scala
oliviertoupin: What about it?

Check MessageMigration.scala in Nrv it's an implicit helper.

mlaflamm: Olivier, FYI I am moving timestamp from message.metadata to message.attachments. This is ok to do the transformation until I merge my stuff.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableContinuousFeeder.scala
oliviertoupin: I agree, and the data doesn't even come from a message (what I expected). I gonna remove.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableContinuousFeeder.scala
```diff
@@ -23,12 +25,13 @@ class TableContinuousFeeder(name: String ,storage: MysqlStorage, table: Table, t
     this.context = context
 
     val data = context.data
-    lastRecord = if (data.contains(Keys) && data.contains(Token) && data.contains(Timestamp))
+    val bool: Boolean = data.containsEither(Timestamp)
```
coderunner: I'm sure this variable can have a better name :)

coderunner:
+1

PR #67, opened 2013-02-13
Ensure GC gauge always use the active TableCollector token value.

mlaflamm:
Only the first gauge is registered to the metrics registery. The gauge must use the active TableCollector token and not the value contained in the TableCollector instance which created the initial gauge.

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
@@ -250,7 +250,7 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
     @volatile
     private var active = false
     private var tokenRanges: List[TokenRange] = List(TokenRange.All)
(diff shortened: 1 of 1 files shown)


coderunner:
+1

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -326,7 +326,13 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
       "gc-extra-record-loaded", table.uniqueName, "extra-record-loaded", TimeUnit.SECONDS))
     private val collectedTokenGauge = metrics.metricsRegistry.newGauge(GarbageCollector.getClass,
       "gc-collected-token", table.uniqueName, new Gauge[Long] {
-        def value = tableNextToken
+        def value = {
+          // Use token from the current table collector
+          GarbageCollector.tableCollectors.get(table) match {
+            case Some(collector) => collector.tableNextToken
+            case _ => 0
```
alexbergeron, line 333: You could replace _ by None, would make it more explicit

PR #68, opened 2013-02-14
Changes for transactions logging

mlaflamm:


mry-core/src/main/scala/com/wajam/mry/Database.scala
```diff
@@ -108,8 +106,7 @@ class Database(var serviceName: String = "database")
 
   private def execute(req: InMessage) {
     var values: Seq[Value] = null
-    val timestamp: Timestamp = req.metadata.getValueOrBuildFromString("timestamp", (v: Seq[String]) => {Timestamp(v(0))}).asInstanceOf[Timestamp]
-    val context = new ExecutionContext(storages, Some(timestamp))
+    val context = new ExecutionContext(storages, Some(Consistency.getMessageTimestamp(req).get))
     context.cluster = Database.this.cluster
```
coderunner, line 110: Shouldn't we think about having timestamp as an attribute in Message one day?

mry-core/src/main/scala/com/wajam/mry/Database.scala
```diff
@@ -117,6 +114,7 @@ class Database(var serviceName: String = "database")
       transaction.execute(context)
       values = context.returnValues
       context.commit()
+      transaction.reset()
```
coderunner, line 118: Why this new line?

mry-core/src/main/scala/com/wajam/mry/Database.scala
mlaflamm, line 110: Totally agree once we are serializing Message with protobuf

coderunner:
+1

PR #74, opened 2013-02-27
Optimize ListValue creation - O(n) instead of O(n^2)

alexbergeron:
Map function on iterator already returns a list. toList doesn't create a new list or do any kind of fancy conversion.

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MultipleRecordValue.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MultipleRecordValue.scala
@@ -33,13 +33,10 @@ class MultipleRecordValue(storage: MysqlStorage, context: ExecutionContext, tabl
     this.iterator match {
       case Some(iter) =>
         try {
-          var list = List[Value]()
-          while (iter.next()) {
-            val record = iter.record
-            list :+= new RecordValue(storage, context, table, token, record.accessPath, Some(transaction), Some(record))
+          val list: Traversable[RecordValue] = iter map {
+            (record) => new RecordValue(storage, context, table, token, record.accessPath, Some(transaction), Some(record))
           }
-
-          new ListValue(list)
+          new ListValue(list.toList)
         } finally {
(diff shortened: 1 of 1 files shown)


mry-core/src/main/scala/com/wajam/mry/storage/mysql/MultipleRecordValue.scala
```diff
@@ -33,13 +33,10 @@ class MultipleRecordValue(storage: MysqlStorage, context: ExecutionContext, tabl
     this.iterator match {
       case Some(iter) =>
         try {
-          var list = List[Value]()
-          while (iter.next()) {
-            val record = iter.record
-            list :+= new RecordValue(storage, context, table, token, record.accessPath, Some(transaction), Some(record))
+          val list: Traversable[RecordValue] = iter map {
+            (record) => new RecordValue(storage, context, table, token, record.accessPath, Some(transaction), Some(record))
```
coderunner, line 37: So this is a good as prepend + reverse?
We can't do prepend without reverse because of offset...let say we ask for offset 0 and count 5 and we get A B C D E from DB but return E D C B A. If then we ask for offset 2 and count 5 we would get G F E D C instead of C B A ...

alexbergeron, line 37: This is as optimal as prepend + reverse, map should iterate only once over the iterator, and it doesn't come up in the profiler

coderunner:
+1

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MultipleRecordValue.scala
appaquet, line 37: The old code is shameful ;) That was my first baby steps in scala :(

alexbergeron, line 37: You had to start somewhere :wink:

PR #75, opened 2013-02-28
MRY refactoring (that doesn't affect performance)

alexbergeron:


suhailshergill:
+1

alexbergeron:
Not merging before Monday, we have more urgent things to release for now...

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
```diff
@@ -44,45 +47,49 @@ class ListValue(var listValue: Seq[Value]) extends Value with Serializable {
   }
 }
 
-class StringValue(var strValue: String) extends Value with Serializable {
+@SerialVersionUID(-3026000576636973393L)
+case class StringValue(strValue: String) extends Value {
   override def toString = strValue
 
   override def equalsValue(that: Value): Boolean = {
```
mlaflamm: I suspect that all the equalsValue override are now useless since that all Value subclass are case classes.

alexbergeron: True, but we do need to keep supporting it for current code. Maybe it should be defined as deprecated in Value, and would only calls equals...

mlaflamm: I am just saying that the implementation in Value is the only one needed since calling == and the subclasses are now case classes.

mlaflamm:
+10,000,000

PR #77, opened 2013-03-01
Fixes timeline feeder loop when most recent timestamp records > batch

### LINKED ISSUE #76: TableTimelineFeeder loop forever if last timestamp has more that 100 records
Loop if the most recent timestamp is a token having more records than batchSize (default=100)

mlaflamm:
Closes #76

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableTimelineFeeder.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableTimelineFeeder.scala
@@ -51,32 +51,34 @@ class TableTimelineFeeder(name: String, storage: MysqlStorage, table: Table, tok
     try {
       transaction = storage.createStorageTransaction
       var mutations = transaction.getTimeline(table, timestampCursor, batchSize, tokenRanges, FromTimestamp)
-      lastSelectMode = FromTimestamp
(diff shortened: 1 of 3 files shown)


mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableTimelineFeeder.scala
```diff
@@ -51,32 +51,34 @@ class TableTimelineFeeder(name: String, storage: MysqlStorage, table: Table, tok
...
-              }
-            })
+      if (!mutations.isEmpty) {
+        if (mutations.head.newTimestamp == mutations.last.newTimestamp) {
+          // The whole batch has the same timestamp and has thus been inserted together in the same transaction.
+          // Reload everything at that timestamp in case some records where drop by the batch size limit.
+          mutations = transaction.getTimeline(table, mutations.head.newTimestamp, 0, tokenRanges, AtTimestamp)
+          lastSelectMode = AtTimestamp
+        } else {
+          lastSelectMode = FromTimestamp
```
mlaflamm, line 61: FYI, there are only 2 changes in this block of code: 1) the if is split into 2 if and 2) this line (i.e. FromTimestamp) is moved from outside the outer if to here.

coderunner:
+1

PR #81, opened 2013-03-04
Fix previous issue #76 fix

mlaflamm:


mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableTimelineFeeder.scala
```diff
@@ -51,34 +51,33 @@ class TableTimelineFeeder(name: String, storage: MysqlStorage, table: Table, tok
...
-                  false
-                } else {
-                  found
-                }
-              })
-            }
-            val after = mutations.size
-            debug("Before {} after {}", before.asInstanceOf[Object], after.asInstanceOf[Object])
+
+      if (mutations.size > 1 && mutations.head.newTimestamp == mutations.last.newTimestamp) {
```
mlaflamm, line 55: mutations.size > 1 is important here. I would have noticed if I had run all the tests for my previous pull request. My bad!!

coderunner:
Can you explain why?

alexbergeron:
I saw that on IRC, but I'm waiting for you to respond to @coderunner before +1ing this

mlaflamm:
Need at least 2 records to verify if all records belongs to the same timestamp to fallback to AtTimestamp mode. This is the same as before my previous fix. 

The real fix of issue #76 is in the else, the feeder must switch back to FromTimestamp mode only if there is at least one record otherwise the feeder will load records from the task context timestamp inclusively and alternating from AtTimestamp to FromTimestamp forever until a new record is inserted (See initialisation of timestampCursor not visible in diff).

alexbergeron:
+1

PR #83, opened 2013-03-04
Extracted consistency related behavior from Database to ConsistentDatabase

mlaflamm:


mry-core/src/main/scala/com/wajam/mry/Database.scala
```diff
@@ -70,18 +68,18 @@ class Database(var serviceName: String = "database")
       }
 
       remoteAction.call(Map(Database.TOKEN_KEY -> context.tokens(0), "trx" -> MMigrationCatchAll(transaction)), // TODO: MigrationDuplicate: Remove trx
-                        data = transaction,
-                        onReply = (resp, optException) => {
-        if (ret != null) {
-          if (optException.isEmpty)
-            if (resp.hasData)
-              ret(resp.getData[Seq[Value]], None)
+        data = transaction,
```
mlaflamm, line 71: Automatic code reformat, nothing have really changed in this block of code.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
```diff
@@ -401,6 +401,37 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
...
+                WHERE tk = ? AND ts = ?;
+                   """.format(fullTableName)
+
+    /*
+     * Generated SQL looks like:
+     *
+     *     DELETE FROM `table2_table2_1_table2_1_1_data`
+     *     WHERE tk = ? AND ts = ?;
+     */
+    val dataSql = """
```
coderunner, line 424: merge two sql into one with table suffix as parameter?

coderunner:
Looks good

+1

PR #85, opened 2013-03-05
Implemented projection for MRY

coderunner:
This implementation is MRY only and not specific to MySql storage. 

Projection is only implemented for data structures (i.e. MapValue and ListValue). Applying projection on anything else will result in an exception.

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
@@ -34,7 +34,11 @@ case class MapValue(mapValue: Map[String, Value]) extends Value {
   }
 
   override def execProjection(context: ExecutionContext, into: Variable, keys: Object*) {
-    into.value = MapValue(mapValue.filter(e => keys.contains(StringValue(e._1))))
+    into.value = if (!keys.isEmpty) {
+      MapValue(mapValue.filter(e => keys.contains(StringValue(e._1))))
+    } else {
+      MapValue(mapValue)
+    }
   }
 }
 
@@ -51,11 +55,14 @@ case class ListValue(listValue: Seq[Value]) extends Value {
   }
 
   override def execProjection(context: ExecutionContext, into: Variable, keys: Object*) {
-    val newList = listValue.map(v => {
-      v.execProjection(context, into, keys: _*)
-      into.value
-    })
-    into.value = ListValue(newList)
+    into.value = if (!keys.isEmpty) {
+      ListValue(listValue.map(v => {
+        v.execProjection(context, into, keys: _*)
+        into.value
+      }))
+    } else {
+      ListValue(listValue)
+    }
   }
 }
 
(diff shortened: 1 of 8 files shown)


mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
```diff
@@ -32,6 +32,10 @@ case class MapValue(mapValue: Map[String, Value]) extends Value {
     val newMap = for ((k, v) <- mapValue) yield (k -> v.serializableValue)
     new MapValue(newMap)
   }
+
+  override def execProjection(context: ExecutionContext, into: Variable, keys: Object*) {
+    into.value = MapValue(mapValue.filter(e => keys.contains(StringValue(e._1))))
```
coderunner: This code assumes that projection keys are StringValue since MapValue is [String, Value].

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
```diff
@@ -45,6 +49,14 @@ case class ListValue(listValue: Seq[Value]) extends Value {
     val newList = for (oldValue <- listValue) yield oldValue.serializableValue
     new ListValue(newList)
   }
+
+  override def execProjection(context: ExecutionContext, into: Variable, keys: Object*) {
+    val newList = listValue.map(v => {
+      v.execProjection(context, into, keys: _*)
+      into.value
```
coderunner: This seems a bit hacky since I'm using the into variable to store temporary projected MapValues but it is the simplest way I found. At the end, the into.value is set to the resulting projected list.

If an exception is thrown, then the into variable would have the last projected MapValue if any.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MultipleRecordValue.scala
```diff
@@ -46,6 +46,9 @@ class MultipleRecordValue(storage: MysqlStorage, context: ExecutionContext, tabl
     }
   }
 
+  // Operations are executed on records (list of map or null)
+  override def proxiedSource: Option[OperationSource] = Some(this.innerValue)
```
coderunner, line 50: I had to add this line so projection would apply to MultipleRecordValue which proxies ListValue.

appaquet:
like it! +1

mry-core/src/main/scala/com/wajam/mry/execution/Variable.scala
```diff
@@ -3,6 +3,7 @@ package com.wajam.mry.execution
 /**
  * Variable that can points to nothing or to a value
  */
+@SerialVersionUID(-2511774283683367275L)
 class Variable(var block: Block, var id: Int, var value: Value = NullValue) extends Object with OperationSource with OperationApi with Serializable {
```
alexbergeron, line 7: Maybe we should check (in a future pull request) if theses can be changed to vals

alexbergeron:
+1

PR #86, opened 2013-03-05
ConsistentStore.getLastTimestamp implementation

mlaflamm:


mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
```diff
@@ -999,4 +999,39 @@ class TestMysqlStorage extends TestMysqlBase {
     after(3) should be(MapValue(Map("k" -> "value3.1")))
     after(4) should be(NullValue)
   }
+
+  test("getLastTimestamp should return max range timestamp") {
+
+    val context = new ExecutionContext(storages)
+    case class Data(key: String, table: String, timestamp: Long = Random.nextLong()) {
+      val token = context.getToken(key)
```
alexbergeron, line 1007: Why did you put that outside of the case class constructor?

alexbergeron:
+1

PR #88, opened 2013-03-11
Read and write mutation messages from and to consistent database

mlaflamm:


mry-core/src/main/scala/com/wajam/mry/storage/mysql/Model.scala
```diff
@@ -30,7 +30,7 @@ trait TableCollection {
   def getTable(name: String): Option[Table] = this.tables.get(name)
 }
 
-class Table(val name: String, var parent: Option[Table] = None, val maxVersions: Int = 3) extends TableCollection {
```
alexbergeron, line 33: :smile_cat:

mry-core/src/main/scala/com/wajam/mry/storage/mysql/Model.scala
```diff
@@ -42,6 +42,13 @@ class Table(val name: String, var parent: Option[Table] = None, val maxVersions:
 
   lazy val uniqueName = depthName("_")
 
+  def path: List[Table] = {
+    this.parentTable match {
+      case None => List(this)
+      case Some(t) => t.path :+  this
```
alexbergeron, line 48: Should the length of this list remain small? If not, this won't scale.

Since I guess the order of the list is important, I'd suggest using a Vector for this.

alexbergeron:
Looks good!

mry-core/src/main/scala/com/wajam/mry/storage/mysql/Model.scala
mlaflamm, line 48: Yes, it should stay small (we don't support more than 3 table levels right now).

alexbergeron:
+1

PR #89, opened 2013-03-11
Removed useless vars in Variable

alexbergeron:


mry-core/src/main/scala/com/wajam/mry/execution/Variable.scala
```diff
@@ -4,7 +4,7 @@ package com.wajam.mry.execution
  * Variable that can points to nothing or to a value
  */
 @SerialVersionUID(-2511774283683367275L)
-class Variable(var block: Block, var id: Int, var value: Value = NullValue) extends Object with OperationSource with OperationApi with Serializable {
+class Variable(val block: Block, val id: Int, var value: Value = NullValue) extends Object with OperationSource with OperationApi with Serializable {
```
mlaflamm: Does block need to be a public val? If not, I would even remove val entirely.

alexbergeron: Doesn't look like it... Testing it right now

mlaflamm:
+1

PR #91, opened 2013-03-14
Use ArrayBuffer to speed up append operation

coderunner:
This is a breaking change as the recipient of the Block will not be able to assign List to ArrayBuffer and vice-versa.

Other option would have been to still use List but prepend element than call reverse before executing, but this is also breaking since the receiver will not have the 'reverse' call and will execute out of order.

This seems to be the safest change to get important (two orders of magnitude) performance improvements on large operation lists.

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/execution/Block.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/execution/Block.scala
@@ -1,13 +1,15 @@
 package com.wajam.mry.execution
 
 import com.wajam.mry.execution.Operation.Return
+import collection.mutable.ArrayBuffer
 
 /**
  * Programmatic block of operations and variables that is executed against storage
  */
 trait Block {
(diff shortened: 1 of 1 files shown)


mry-core/src/main/scala/com/wajam/mry/execution/Block.scala
```diff
@@ -1,13 +1,15 @@
...
 
 /**
  * Programmatic block of operations and variables that is executed against storage
  */
 trait Block {
-  var variables = List[Variable]()
-  var operations = List[Operation]()
+  //TODO should not expose these publicly, wait til Protobuf MRY codec is done
+  var variables = ArrayBuffer[Variable]()
+  var operations = ArrayBuffer[Operation]()
```
alexbergeron, line 12: Could it be vals?

alexbergeron:
Add a todo to change vars to vals then +1

PR #94, opened 2013-03-19
Rollback storage transaction if execution time > database service response timeout

mlaflamm:
Before the switchboard was timing out but the MRY transaction wasn't rolled back which produced inconsistencies between the transaction log and the database.

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/Database.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/Database.scala
@@ -7,14 +7,14 @@ import com.yammer.metrics.scala.Instrumented
 import com.wajam.nrv.service._
 import com.wajam.nrv.tracing.Traced
 import com.wajam.nrv.data.InMessage
-import com.wajam.nrv.utils.{Promise, Future}
+import com.wajam.nrv.utils.{CurrentTime, Promise, Future}
 import com.wajam.nrv.consistency.Consistency
 
 /**
  * MRY database
  */
 class Database[T <: Storage](serviceName: String = "database")
-  extends Service(serviceName) with Logging with Instrumented with Traced {
+  extends Service(serviceName) with CurrentTime with Logging with Instrumented with Traced {
 
   var storages = Map[String, T]()
 
@@ -105,7 +105,7 @@ class Database[T <: Storage](serviceName: String = "database")
   }, ActionMethod.GET))
   remoteReadExecuteToken.applySupport(resolver = Some(Database.TOKEN_RESOLVER))
 
-  private def transactionTimeout = math.max(responseTimeout * 0.9, responseTimeout - 1000)
+  private def transactionTimeout = math.max(responseTimeout * 0.75, responseTimeout - 500)
 
   private def execute(req: InMessage) {
     var values: Seq[Value] = null
@@ -115,9 +115,9 @@ class Database[T <: Storage](serviceName: String = "database")
     try {
       val transaction = req.getData[Transaction]
 
-      val startTime = System.currentTimeMillis()
+      val startTime = currentTime
       transaction.execute(context)
-      val elapsedTime = System.currentTimeMillis() - startTime
+      val elapsedTime = currentTime - startTime
       if (elapsedTime > transactionTimeout) {
         throw new TimeoutException("Database transaction took too much time to execute", Some(elapsedTime))
       }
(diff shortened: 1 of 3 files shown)


mry-core/src/main/scala/com/wajam/mry/Database.scala
```diff
@@ -105,6 +105,8 @@ class Database[T <: Storage](serviceName: String = "database")
   }, ActionMethod.GET))
   remoteReadExecuteToken.applySupport(resolver = Some(Database.TOKEN_RESOLVER))
 
+  private def transactionTimeout = math.max(responseTimeout * 0.9, responseTimeout - 1000)
```
alexbergeron: How was that determined?

mlaflamm: Spiting in the air and making sure there is just enough wind to miss myself.

alexbergeron: On a calm day without wind, what would happen?

mlaflamm: Magik, Magiiiik! (to say with an east European accent).

mlaflamm: Seriously the database timeout need to be shorter than the switchboard timeout otherwise if this is too close, the database may be successful but the switchboard may timeout milliseconds before receiving the response. I tough that one second shorter is very safe and acceptable up to a timeout of as small than 4-5 seconds. For timeout smaller than that, the value must be less than a second. I am not sure this should really be configurable as this might be error prone but I was expecting comments on my selection of magic numbers... 

You just made me realize that 0.75 and 500 should probably a better magic numbers...

alexbergeron: Ooooooooo :open_mouth: 

Yeah, make sense. I trust you.

My comment were kinda aiming at how arbitrary you chose your numbers, not to extract them to constants. Your logic makes sense.

alexbergeron:
+1

mry-core/src/main/scala/com/wajam/mry/Database.scala
```diff
@@ -113,7 +115,12 @@ class Database[T <: Storage](serviceName: String = "database")
     try {
       val transaction = req.getData[Transaction]
 
+      val startTime = System.currentTimeMillis()
```
coderunner: Don't we have a TimeGenerator trait somewhere? That would simplify testing I think.

Or make something like...

``` scala
trait CurrentTime {
  def timeAccessor: TimeAccessor

  trait TimeAccessor { def currentTime: Long }
}

// real
trait SystemTime extends CurrentTime {
  val timeAccessor = new TimeAccessor { def currentTime = System.currentMillis }
}

// test
val mockTime = mock[CurrentTime.TimeAccessor]
trait MockTime extends CurrentTime {
  val timeAccessor = mockTime
}

abstract class Database with CurrentTime  {
...}

val db = new Database with SystemTime

val testDB = new Database with MockTime
```

mry-core/src/test/scala/com/wajam/mry/TestDatabase.scala
```diff
@@ -83,4 +88,56 @@ class TestDatabase extends FunSuite with BeforeAndAfterAll {
...
+      }, set1.complete(_, _))
+      Future.blocking(set1.future, 5000)
+      val Some(Right(Seq(set1Value))) = set1.future.value
+      assert(set1Value.equalsValue("value1"))
+
+      // Add a artificial delay greater than the service response timeout to storage execution.
+      val spyStorage = db.getStorage("memory").asInstanceOf[MemoryStorage]
+      when(spyStorage.createStorageTransaction(anyObject())).thenAnswer(new Answer[spyStorage.MemoryTransaction]() {
+        def answer(invocation: InvocationOnMock) = {
+          Thread.sleep(db.responseTimeout * 2)
```
coderunner, line 114: we should try to remove sleep from tests.

mlaflamm, line 114: I know, I know. This is more complicated to mock the time because I need to do synchronize it between multiple components (e.g. the switchboard). I also wanted to see the real interaction between the various components that are also sensitive to time.

coderunner:
Change the transactionTimeout then +1

PR #95, opened 2013-03-19
Exclude read transactions iterator records from GC

mlaflamm:
Also setup a function to consistent store to lookup the current consistent timestamp.

### DIFF
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
@@ -543,7 +543,6 @@ class TestMysqlStorage extends TestMysqlBase {
     val token1 = context.getToken("key1")
     val token2 = context.getToken("key2")
     val ranges = List(TokenRange(token1, token1), TokenRange(token2, token2))
-    mysqlStorage.setLastConsistentTimestamp(Long.MaxValue, ranges)
     val table1RangeTimeline = mysqlStorage.createStorageTransaction(context).getTimeline(table1, createTimestamp(0), 100, ranges)
     val allKeys = table1Timeline.map(_.accessPath.toString)
     val expectedRangeKeys = table1Timeline.filter(r => token1 == r.token || token2 == r.token).map(_.accessPath.toString)
@@ -785,7 +784,6 @@ class TestMysqlStorage extends TestMysqlBase {
     }
 
     val ranges = List(TokenRange(1000000001L, 2000000000L), TokenRange(3000000001L, 4000000000L))
-    mysqlStorage.setLastConsistentTimestamp(Long.MaxValue, ranges)
     mysqlStorage.GarbageCollector.setCollectedRanges(ranges)
 
     // Create 5 versions for each token/keys tupple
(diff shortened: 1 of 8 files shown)


mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
```diff
@@ -844,19 +842,37 @@ class TestMysqlStorage extends TestMysqlBase {
...
+    exec(_.from("mysql").from("table1").set("k", Map("k" -> "1")), commit = true, onTimestamp = 100L)
+    exec(_.from("mysql").from("table1").set("k", Map("k" -> "1")), commit = true, onTimestamp = 200L)
+    exec(_.from("mysql").from("table1").set("k", Map("k" -> "1")), commit = true, onTimestamp = 300L)
+    exec(_.from("mysql").from("table1").set("k", Map("k" -> "1")), commit = true, onTimestamp = 400L)
+    exec(_.from("mysql").from("table1").set("k", Map("k" -> "1")), commit = true, onTimestamp = 500L)
+    exec(_.from("mysql").from("table1").set("k", Map("k" -> "1")), commit = true, onTimestamp = 600L)
+
+    // Force collection, versions from the open transactions iterator start timestamp are ignored
+    val itr = mysqlStorage.readTransactions(400L, 600L, Seq(TokenRange.All))
+    mysqlStorage.GarbageCollector.collectAll(100) should be(1)
```
alexbergeron, line 872: If this fails, iterator won't be closed

mlaflamm, line 872: Are you talking about the test or the design? The storage is discarded at the end of the test. When used via ConsistentDatabase (i.e. real world usage), the iterator is close in case of error.

alexbergeron, line 872: Talking about the test. It's mostly to be sure that even if it fails, the storage has no problem with an open iterator.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -756,6 +762,15 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
       debug("GCing iteration done on {}! Collected {} versions", table.uniqueName, collectedTotal)
       collectedTotal
     }
+
+    /**
+     * Returns the consistent timestamp that must be used by GC. The consistent timestamp exclude the transactions
+     * currently read for replication.
+     */
+    private def getGcConsistentTimestamp(ranges: Seq[TokenRange]): Timestamp = {
+      val rangesReadStartTimestamps = openReadIterators.withFilter(!_.ranges.intersect(ranges).isEmpty).map(_.from)
```
alexbergeron, line 771: Why withFilter? Why not just filter?

mlaflamm, line 771: I use mostly <b>withFilter</b> whenever I can for single traversal usage. The <b>filter</b> method creates a new filtered collection in place while <b>withFilter</b> is a view that wraps the filtered collection and that is applied lazily as you traverse the collection (here triggered by <b>map</b>). This is great for very large collection but probably not very useful here. I did it more out of habit than anything else.

coderunner:
+1

PR #97, opened 2013-03-27
Transaction's protobuf codec

oliviertoupin:


alexbergeron:
Is it compatible with data already stored in MRY?

mry-core/src/main/scala/com/wajam/mry/api/TransactionPrinter.scala
```diff
@@ -0,0 +1,76 @@
...
+  * Generate a very detailed pretty print of a Transaction tree.
+  * Used for debugging.
+  *
+  */
+object TransactionPrinter {
+
+  def printTree(transaction: Transaction, contextName: String): String = {
+    val builder = new StringBuilder()
+
+    builder append "--%s--\n".format(contextName)
```
alexbergeron, line 16: This is nitpicking, but I'm not a fan of this kind of notation on StringBuilder.

See http://docs.scala-lang.org/style/method-invocation.html#arity1

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,451 @@
...
+          case pMryData if pMryData.hasTransaction => pMryData.getTransaction
+          case pMryData if pMryData.hasBlock => pMryData.getBlock
+          case pMryData if pMryData.hasVariable => pMryData.getVariable
+          case pMryData if pMryData.hasOperation => pMryData.getOperation
+          case pMryData if pMryData.hasValue => pMryData.getValue
+        }
+
+        tempDecodingHeap += (addr -> value)
+        addr += 1
+    }
```
alexbergeron, line 153: Appending to a TreeMap is not the most efficient operation. 
Also, your indexes are all numbers from 1 to the length of your value list...

Are you _really_ sure you want to do it this way? AFAIK, a sequence would be fine (you could always zip it with a range going from 1 to the size of the sequence)

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,451 @@
...
+  }
+
+  private def getFromHeap[T](address: Int): T = {
+
+    checkAddress(address)
+
+    tempDecodingHeap(address).asInstanceOf[T]
+  }
+
+  private def loadHeap(pDecodingHeap: PHeap) = {
```
alexbergeron, line 137: Remove =

Or make this function return something usefull.

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,451 @@
...
+  private def decodeBlockFromHeapToPool(block: Block) = {
+
+    // Yield map with address
+
+    val mappedData = tempDecodingHeap
+
+    // Decode values and register them to pool
+    val values = mappedData
+      .filter(_._2.isInstanceOf[PTransactionValue])
+      .map((v) => v._1 -> decodePValue(v._2.asInstanceOf[PTransactionValue]))
```
alexbergeron:
``` scala
val values = mappedData collect {
  case (key, pValue: PTransactionValue) => (key -> decodePValue(pValue))
}
```

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,451 @@
...
+
+    values.foreach((v) => registerDecodedData(v._1 , v._2))
+
+    // Decode variables and register them to pool
+    val variables = mappedData
+      .filter(_._2.isInstanceOf[PVariable])
+      .map((v) => v._1 -> decodePVariable(block, v._2.asInstanceOf[PVariable]))
```
alexbergeron: Same pattern

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,451 @@
...
+
+    variables.foreach((v) => registerDecodedData(v._1, v._2))
+
+    // Decode operations and register them to pool
+    val operations = mappedData
+      .filter(_._2.isInstanceOf[POperation])
+      .map((v) => v._1 -> decodePOperation(block, v._2.asInstanceOf[POperation]))
```
alexbergeron: Same pattern

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,451 @@
...
+
+    loadHeap(transport.getHeap)
+
+    val request: Option[Transaction] =
+      if (transport.hasRequestAddress)
+      {
+        Some(decodePTransaction(transport))
+      }
+      else
+        None
```
alexbergeron, line 207: Inconsistent style between this if and the rest of the code...

Also, inconsistent style between this if and itself :smirk:

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,451 @@
...
+
+    val response: Seq[Value] =
+      transport.getResponseAddressesList
+        .map(getFromHeap[PTransactionValue](_))
+        .map(decodePValue(_)).toSeq
```
alexbergeron: Looks like a good case for function composition...

``` scala
transport.getResponseAddressesList map (getFromHeap[PTransactionValue] _ andThen decodePValue _) toSeq
```

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,451 @@
...
+
+  private def encodePBlock(block: Block): PBlock.Builder =  {
+    val pBlock = PBlock.newBuilder()
+
+    val varId = block.variables.map(encodePVariable(_))
+
+    varId.zipWithIndex.foreach((v) => registerEncodedData(v._1, block.variables(v._2)))
+    pBlock.addAllVariableAddresses(varId.map(_.asInstanceOf[java.lang.Integer]))
+
+    pBlock.addAllOperationAddresses(block.operations.map(encodePOperation(_)).map(addToHeap(_)))
```
alexbergeron:
``` scala
pBlock.addAllOperationAddresses(block.operations.map(encodePOperation _ andThen addToHeap _))
```

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,451 @@
...
+
+    import Operation._
+
+    val pOperation = POperation.newBuilder()
+
+    pOperation.setSourceAddress(getEncodedId(operation.source))
+
+    val withFrom = (op: WithFrom) =>
+      // Variable are already listed in the parent block, so don't duplicate
+      pOperation.addAllVariableAddresses(op.from.map(getEncodedId(_)).map(_.asInstanceOf[java.lang.Integer]))
```
alexbergeron:
``` scala
pOperation.addAllVariableAddresses(op.from.map(getEncodedId(_).asInstanceOf[java.lang.Integer]))
```

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,451 @@
...
+
+    val WithIntoAndSeqObject = (into: Variable, objects: Seq[Object]) =>
+      pOperation
+        .addVariableAddresses(getEncodedId(into)) // Variable are already listed in the parent block, so don't duplicate
+        .addAllObjectAddresses(objects.map(encodePObject(_)).map(_.asInstanceOf[java.lang.Integer]))
```
alexbergeron: Same here.

Why are you calling map twice so often?

oliviertoupin:
Correct me if I'm wrong but Mry store a Value in the blob, and this part
wasn't modified, so should remain compatible. MryStorage use encodeValue,
decodeValue

Look @ /com/wajam/mry/storage/mysql/MysqlTransaction.scala:102

alexbergeron:
Yeah, I'm realizing that you kept compatibility. You should still test it with actual data (let's say you could use sandbox as you local mysql db)

mry-core/src/main/scala/com/wajam/mry/execution/Block.scala
```diff
@@ -49,4 +50,37 @@ trait Block {
...
+    for (op <- operations)
+      if (op._1.source eq t1)
+      {
+        assert(op._2.source eq t2)
+      }
+      else
+      {
+        val vs = variables.find((v) => op._1.source eq v._1).get
+        assert(op._2.source eq vs._2)
+      }
```
alexbergeron: Pattern matching is your friend :smile: 

``` scala
for ((op1, op2) <- operations) {
  if (op1.source eq t1) {
    assert(op2.source eq t2
  } else {
    val (_, vs) = variables.find((v) => op1.source eq v._1).get
    assert(op2.source eq vs)
  }
}
```

Also, .get will crash if not found (but I think that's what you're expecting since you're doing assertions)

mry-core/src/main/scala/com/wajam/mry/execution/Operation.scala
```diff
@@ -1,67 +1,144 @@
...
+  class Get(source: OperationSource, val into: Variable, val keys: Object*) extends Operation(source) with Serializable {
     def execute(context: ExecutionContext) {
       source.execGet(context, into, keys: _*)
     }
 
+    override def equalsContent(obj: Any): Boolean = {
+      obj match {
+        case o: Get =>
+          into.equalsContent(o.into) &&
+          keys.forall((k) => o.keys.exists(k.equalsContent(_)))
```
alexbergeron: I'm not a fan of this indentation here... Could you put it in brackets?

mry-core/src/main/scala/com/wajam/mry/execution/Variable.scala
```diff
@@ -12,4 +14,14 @@ class Variable(block: Block, id: Int, var value: Value = NullValue) extends Obje
   def reset() {
     this.value = NullValue
   }
+
+  override def equalsContent(obj: Any): Boolean = {
+    obj match {
+      case v: Variable =>
+        value.equalsValue(v.value) &&
+        this.id == v.id
```
alexbergeron: Why is this
on two lines?

mry-core/src/test/scala/com/wajam/mry/api/protobuf/TestProtobufTranslator.scala
```diff
@@ -4,13 +4,19 @@ import org.scalatest.FunSuite
 import com.wajam.mry.execution.Implicits._
 import org.junit.runner.RunWith
 import org.scalatest.junit.JUnitRunner
-import com.wajam.mry.execution.{ListValue, Value, MapValue, NullValue}
+import com.wajam.mry.execution._
+import com.wajam.mry.execution.MapValue
+import com.wajam.mry.execution.ListValue
```
alexbergeron: You don't need to import MapValue and ListValue explicitly

mry-core/src/test/scala/com/wajam/mry/api/protobuf/TestProtobufTranslator.scala
```diff
@@ -65,8 +71,163 @@ class TestProtobufTranslator extends FunSuite {
     }
   }
 
+  private def buildTransaction(): Transaction = {
+    val t = new Transaction((b) => b.returns(b.from("B").get(1000).set("C").limit(100).projection("D").delete("E")))
```
alexbergeron: You really need to assign it to a val?

mry-core/src/test/scala/com/wajam/mry/api/protobuf/TestProtobufTranslator.scala
```diff
@@ -65,8 +71,163 @@ class TestProtobufTranslator extends FunSuite {
...
+
+    val results2 = transport2.response
+
+    transport2.request.isDefined should be(false)
+
+    val merged = results.zip(results2)
+
+    merged.map { (zip) =>
+      zip._1 equalsValue zip._2 should be(true)
+    }
```
alexbergeron, line 174:
``` scala
merged map {
  case (z1, z2) => (z1 equalsValue z2) should equal (true)
}
```

alexbergeron, line 174: If you're even more motivated, you could implement your own matcher:
http://www.scalatest.org/user_guide/using_matchers#usingCustomMatchers

mry-core/src/test/scala/com/wajam/mry/api/protobuf/TestProtobufTranslator.scala
```diff
@@ -65,8 +71,163 @@ class TestProtobufTranslator extends FunSuite {
...
+  /**
+   * Check that all operation.source from t1
+   * match the equivalent source from t2
+   */
+  def validateOperationsSources(t1: Transaction, t2: Transaction) {
+
+    val operations = t1.operations.zip(t2.operations)
+    val variables = t1.variables.zip(t2.variables)
+
+    for (op <- operations)
```
alexbergeron:
``` scala
for ((op1, op2) <- operations)
```

mry-core/src/test/scala/com/wajam/mry/api/protobuf/TestProtobufTranslator.scala
```diff
@@ -65,8 +71,163 @@ class TestProtobufTranslator extends FunSuite {
...
+   * Check that all operation variables are bounds to the a variable
+   * from block and not a copy
+   */
+  def validateOperationVariablesAreBoundsToBlock(t: Transaction) {
+
+    for (o <- t.operations) {
+      o match {
+        case _: Return =>
+
+          val op = o.asInstanceOf[WithFrom]
```
alexbergeron: My question yesterday is: why do you need asInstanceOf[WithFrom]?

You already have access to from here!

mry-core/src/test/scala/com/wajam/mry/api/protobuf/TestProtobufTranslator.scala
```diff
@@ -65,8 +71,163 @@ class TestProtobufTranslator extends FunSuite {
...
+          assert(op.from.forall((v) => t.variables.exists(v eq _)))
+
+        case _: From  |
+             _: Get |
+             _: Projection |
+             _: Limit =>
+
+          val op = o.asInstanceOf[WithIntoAndKeys]
+          assert(Some(op.into).forall((v) => t.variables.exists(v eq _)))
```
alexbergeron: Now that I can read it, I see that this line doesn't even _really_ fit with the style of the other lines here IMHO.

``` scala
assert(t.variables.exists(_ eq op.into))
```

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
oliviertoupin, line 153: I'm double checking, but I had a problem with ordering before, so I switched to a TreeMap

alexbergeron, line 153: If it's in a sequence, you wouldn't have this problem.

mry-core/src/test/scala/com/wajam/mry/api/protobuf/TestProtobufTranslator.scala
oliviertoupin: Mosty to fit style with the next statement, where we can't name the variable in the case.

mry-core/src/main/protobuf/MryProtobuf.proto
```diff
@@ -0,0 +1,112 @@
...
+	optional PTransactionValue value = 5;
+}
+
+/* Transport */
+
+message PTransport {
+
+	optional PHeap heap = 3;
+
+	optional uint32 /* <PTransaction> */ requestAddress = 1;
```
coderunner: why use address? this refer to the id no?

mry-core/src/main/scala/com/wajam/mry/api/MryCodec.scala
```diff
@@ -0,0 +1,34 @@
...
+import protobuf.ProtobufTranslator
+import com.wajam.mry.execution.{Value, Transaction}
+
+class MryCodec extends Codec {
+
+  def encode(entity: Any, context: Any = null): Array[Byte] = {
+
+    val transport = entity match {
+      case request: Transaction => Transport(Some(request), Seq())
+      case response: Seq[Value] => Transport(None, response)
```
coderunner: don't forget type erasure

mry-core/src/main/scala/com/wajam/mry/api/HydridCodec.scala
```diff
@@ -0,0 +1,78 @@
+package com.wajam.mry.api
+
+import com.wajam.nrv.protocol.codec.GenericJavaSerializeCodec
+import com.wajam.nrv.protocol.codec.Codec
+import com.wajam.nrv.Logging
+import java.nio.ByteBuffer
+
+class HybridCodec(mode: HybridCodec.TranslationMode.Value) extends Codec with Logging {
```
coderunner: TransitionCode? MigrationCode?

```
@deprecated?
```

mry-core/src/main/scala/com/wajam/mry/api/HydridCodec.scala
```diff
@@ -0,0 +1,78 @@
...
+  private val mryCodec = new MryCodec
+  private val genericCodec = new GenericJavaSerializeCodec
+
+  // Source: http://docs.oracle.com/javase/6/docs/platform/serialization/spec/protocol.html
+  private val JavaSerializeMagicShort : Short = (0xACED).toShort
+
+  object TranslationMode extends Enumeration {
+
+    // Decode Both, Encode Java, used as a fallback, or a as migration first step
+    val BothThenJava = Value(1)
```
coderunner: DecBothEncJava? I think it clearer than using 'then'

mry-core/src/main/scala/com/wajam/mry/api/HydridCodec.scala
```diff
@@ -0,0 +1,78 @@
...
+
+object HybridCodec {
+
+  private val mryCodec = new MryCodec
+  private val genericCodec = new GenericJavaSerializeCodec
+
+  // Source: http://docs.oracle.com/javase/6/docs/platform/serialization/spec/protocol.html
+  private val JavaSerializeMagicShort : Short = (0xACED).toShort
+
+  object TranslationMode extends Enumeration {
```
coderunner: TransitionMode?

mry-core/src/main/scala/com/wajam/mry/api/Transport.scala
```diff
@@ -0,0 +1,6 @@
+package com.wajam.mry.api
+
+import com.wajam.mry.execution.{Transaction, Value}
+
+case class Transport(request: Option[Transaction], response: Seq[Value]) {
+}
```
coderunner: you can remove empty {}

Shouldn't it be Option[Seq[Value]] to be consistent?

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,446 @@
...
+ * Allow stateful and threadsafe operation for ProtobufTranslator by instancing this class at every call.
+ * It also keep ProtobufTranslator class more readable and more relevant to it's parent interface.
+ *
+ */
+private class InternalProtobufTranslator {
+
+  private var currentAddress = 1 // I don't think will have more than 2^31 object, skip 0, since 0==unassigned
+
+  private val pb2obj = new collection.mutable.HashMap[Int, AnyRef] // Encoded address to decoded instance mapping
+  private val obj2pb = new collection.mutable.HashMap[AnyRef, Int] // Live instance to encoded address mapping
```
coderunner: Are you sure every object has proper support for equals/hashcode?

Can't we restrict to Object?

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,446 @@
...
+
+  private def registerEncodedData(pbAddress: Int, instance: AnyRef) = {
+    obj2pb += instance -> pbAddress
+  }
+
+  private def registerDecodedData(pbAddress: Int, instance: AnyRef) = {
+    pb2obj += pbAddress -> instance
+  }
+
+  private def getEncodedId(instance: AnyRef): Int = {
```
coderunner: again...Id and address mixed up...pick one :) I prefer id, it's less confusing.

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,446 @@
...
+      case value: PBlock.Builder => pMryData.setBlock(value)
+      case value: PVariable.Builder => pMryData.setVariable(value)
+      case value: POperation.Builder => pMryData.setOperation(value)
+      case value: PTransactionValue.Builder => pMryData.setValue(value)
+    }
+
+    pMryData
+  }
+
+  private def addToHeap(mryData: AnyRef): Int = {
```
coderunner, line 113: not sure I like the Heap analogy...I find it confusing too.

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,446 @@
...
+          case pMryData if pMryData.hasOperation => pMryData.getOperation
+          case pMryData if pMryData.hasValue => pMryData.getValue
+        }
+
+        tempDecodingHeap = tempDecodingHeap :+ value
+        addr += 1
+    }
+  }
+
+  private def decodeBlockFromHeapToPool(block: Block) = {
```
coderunner, line 156: why not just use Pool for both concepts?

mry-core/src/main/scala/com/wajam/mry/execution/Block.scala
```diff
@@ -49,4 +50,37 @@ trait Block {
...
+    }
+  }
+}
+
+object Block {
+
+  /**
+   * Check that all operation.source match
+   */
+  protected[mry] def validateOperationsSources(t1: Transaction, t2: Transaction) {
```
coderunner: test method only, should be in tests

mry-core/src/main/scala/com/wajam/mry/execution/Operation.scala
```diff
@@ -1,67 +1,145 @@
...
  */
-abstract class Operation(var source: OperationSource) extends Executable with Serializable {
+abstract class Operation(var source: OperationSource) extends Executable with ContentEquals with Serializable {
 }
 
 object Operation {
 
-  class Return(source: OperationSource, from: Seq[Variable]) extends Operation(source) with Serializable {
+  // TODO: To be replaced by proper hierachy when Java serialization in no longer used.
+  type WithIntoAndData = {def into: Variable; def data: Seq[Object]}
```
coderunner: not sure why we need these?

mry-core/src/main/protobuf/MryProtobuf.proto
oliviertoupin: It refers to auto generated id. But there is also ids in SOME of the class we serialize, so to avoid confusion I use address. I also fit better with the memory model comparaison I use in the comments. 

However it does refers to the position in PHeap, so an alternative name could be "Pos"

mry-core/src/main/scala/com/wajam/mry/api/TransactionPrinter.scala
oliviertoupin, line 16: I do like it, but if you really want me to change it, I'll change it.

oliviertoupin:
All is fixed!

mry-core/src/main/scala/com/wajam/mry/api/Transport.scala
oliviertoupin: It impossible to differenciate between None and empty Seq in the schema we have, so that's why it's like that, I just remembered.

oliviertoupin:
"Yeah, I'm realizing that you kept compatibility. You should still test it with actual data (let's say you could use sandbox as you local mysql db)"

I just did that, but we can't test mry codec that way (sandbox is not aware), however I did find java seralization issue doing that, and fixed those.

alexbergeron:
> I did find java seralization issue doing that, and fixed those.

That's what I wanted you to test :smile:

oliviertoupin:
Perfect then! Anything else before +1

coderunner:
Should be good to go! merge and deploy tomorrow morning?

oliviertoupin:
Exactly!

mry-core/src/main/scala/com/wajam/mry/Database.scala
```diff
@@ -1,5 +1,7 @@
 package com.wajam.mry
 
+import api.HybridCodec.TransitionMode
+import api.{HybridCodec}
```
alexbergeron: You don't need to import HybridCodec twice

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -1,79 +1,421 @@
 package com.wajam.mry.api.protobuf
 
 import com.wajam.mry.api.{TranslationException, ProtocolTranslator}
-import com.wajam.mry.api.protobuf.Transaction.{PTransactionCollectionValue, PTransactionCollection, PTransactionValue}
 import scala.collection.JavaConversions._
-import com.wajam.mry.execution.{Transaction => MryTransaction, _}
+import com.wajam.mry.execution._
+import com.wajam.mry.execution.Transaction
```
alexbergeron: Already imported by import com.wajam.mry.execution._

alexbergeron:
Clean-up imports then +1!

PR #101, opened 2013-04-03
Fix production problem (Nulls, and percolation)

oliviertoupin:


mry-core/src/main/protobuf/MryProtobuf.proto
```diff
@@ -32,6 +32,14 @@ message PMryData {
 
 message PTransport {
 
+    enum Type {
```
coderunner, line 35: Indent seems wrong. Also, should it be EmptyResponse instead of just empty? The semantic seems weird to have three options, it is either a request or a response. It is either empty or not.

mry-core/src/main/scala/com/wajam/mry/Database.scala
```diff
@@ -110,6 +108,9 @@ class Database[T <: Storage](serviceName: String = "database")
   }, ActionMethod.GET))
   remoteReadExecuteToken.applySupport(resolver = Some(Database.TOKEN_RESOLVER))
 
+  remoteWriteExecuteToken.applySupport(nrvCodec = Some(new HybridCodec(TransitionMode.DecBothEncMry)))
```
coderunner, line 112: Shouldn't we put those lines right after the action definition?

oliviertoupin, line 112: I found it more readable that way, plus with default formating, on an other line, this line get indented, which is very annoying

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -177,18 +177,26 @@ private class InternalProtobufTranslator {
...
-    val Transport(request, response) = transport
+
+    import PTransport.Type
+
+    val Transport(transaction, values) = transport
 
     val pTransport = PTransport.newBuilder()
 
-    for (r <- request) {
+    if (transaction.isEmpty && values.isEmpty)
```
coderunner: pattern matching

coderunner:
+1

PR #104, opened 2013-04-16
Use Mry protobuf codec for the whole Database service

mlaflamm:


mry-core/src/main/scala/com/wajam/mry/ConsistentDatabase.scala
```diff
@@ -68,6 +68,11 @@ class ConsistentDatabase[T <: ConsistentStorage](serviceName: String = "database
           val transaction = new Transaction()
           value.applyTo(transaction)
           val message = new InMessage(Map(Database.TOKEN_KEY -> value.token), data = transaction)
+          message.token = value.token
+          message.serviceName = name
+          message.method = ActionMethod.POST
+          message.function = MessageType.FUNCTION_CALL
+          message.path = remoteWriteExecuteToken.path.buildPath(message.parameters)
```
oliviertoupin, line 75: Why those additions?

mlaflamm, line 75: I am recreating MRY write messages from the database data. The most important thing is fill in the message data with the Transaction (was obviously already done). This is not critical to also include the metadata but this helps debugging very much.

oliviertoupin, line 75: ok

oliviertoupin:
+1

PR #106, opened 2013-04-17
Greatly improved performance of replication selects in mysql

mlaflamm:
Now use proper indexes...

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
@@ -479,14 +479,14 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
 
     private def loadCache() {
       try {
-        debug("loadCache - start")
+        trace("loadCache - start")
         // Load more mutation groups if needed and possible
         if (groupsQueue.isEmpty && (!groupsSummaryQueue.isEmpty ||
           tablesCache.exists(ti => ti.hasMore || !ti.records.isEmpty))) {
 
           // Ensure we have a minimum quantity of transaction summary loaded from each table
           tablesCache.withFilter(_.mustLoadMore).foreach(loadTableSummary(_))
-          debug("loadCache - loadedTableSummary done")
+          trace("loadCache - loadedTableSummary done")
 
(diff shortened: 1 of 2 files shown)


mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -29,6 +29,9 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
   private var openReadIterators: List[MutationGroupIterator] = List()
```
mlaflamm, line 29: Changes in this file are only new timers and new logging.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -544,9 +557,12 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
...
         val grouped: mutable.Map[Timestamp, MutationGroup] = mutable.Map()
-        for (table <- tables; record <- trx.getTransactionRecords(table, from, to, ranges)) {
-          val group = grouped.getOrElseUpdate(record.timestamp, MutationGroup(record.token, record.timestamp))
-          group.records = record :: group.records
+        for (table <- tables) {
+          trace("loadTablesData: {}", table)
+          for (record <- trx.getTransactionRecords(table, from, to, ranges)) {
+            val group = grouped.getOrElseUpdate(record.timestamp, MutationGroup(record.token, record.timestamp))
+            group.records = record :: group.records
+          }
```
alexbergeron, line 565: If you want to do this faster

``` scala
tables foreach (table =>  {
  trace("loadTablesData: {}", table)
  trx.getTransactionRecords(table, from, to, ranges)) foreach (record => {
    val group = grouped.getOrElseUpdate(record.timestamp, MutationGroup(record.token, record.timestamp))
    group.records = record :: group.records
  })
})
```

mlaflamm, line 565: The optimisation is not in this file Alex! :stuck_out_tongue_winking_eye: I've just split the for in 2 to add a trace...

coderunner:
+1

PR #110, opened 2013-04-23
More optimal timeline query

appaquet:


coderunner:
Looks good +1. Manu will also take a look.

mlaflamm:
+1. I like the simplification of the query!

PR #112, opened 2013-04-24
Support multi keys operations in TableValue to simplify replicated transactions

mlaflamm:
Prevent errors when parent record does not exist at the replicated timestamp (i.e. when all parent records before that timestamp removed by GC).

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
@@ -242,21 +242,24 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
 
   override def execFrom(context: ExecutionContext, into: Variable, keys: Object*) {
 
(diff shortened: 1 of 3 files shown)


oliviertoupin:
As a general note, this file is getting big, and contains several non-trivial class that could benefit from their on file.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -241,16 +241,38 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
...
 
   override def execFrom(context: ExecutionContext, into: Variable, keys: Object*) {
-    val tableName = param[StringValue](keys, 0).strValue
-    val optTable = model.getTable(tableName)
 
-    optTable match {
-      case Some(t) =>
-        into.value = new TableValue(this, t)
+    def getTable(coll: TableCollection, tableNames: List[String]): Table = {
+      if (tableNames == Nil) {
```
coderunner: Change following our discussion.

coderunner:
+1

PR #113, opened 2013-04-25
Filtering support in Mry

oliviertoupin:
Unit test in TestOperations (Test the filtering directly on MapValue and ListValue)
IT test in TestMySqlStorage (Test the protobuf transport, and the new operation)

### DIFF
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
@@ -63,7 +63,8 @@ class TestMysqlStorage extends TestMysqlBase with ShouldMatchers {
       t.ret(table.get("key1").filter("A", MryFilters.Equals, "2"))
     }, commit = false)
 
-    val finalList = v.value.asInstanceOf[MapValue].mapValue("list").asInstanceOf[ListValue]
+    val MapValue(mapValue) = v.value
+    val ListValue(finalList) = mapValue("list")
 
     finalList.size should equal(2)
 
@@ -92,7 +93,8 @@ class TestMysqlStorage extends TestMysqlBase with ShouldMatchers {
       t.ret(table.get("key1").filter("A", MryFilters.LesserThanOrEqual, 3))
     }, commit = false)
 
-    val finalList = v.value.asInstanceOf[MapValue].mapValue("list").asInstanceOf[ListValue]
+    val MapValue(mapValue) = v.value
+    val ListValue(finalList) = mapValue("list")
 
     finalList.size should equal(2)
 
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
@@ -71,6 +71,35 @@ class TestMysqlStorage extends TestMysqlBase with ShouldMatchers {
     finalList(1) should be(MapValue(Map("A" -> toVal("2"))))
   }
 
+  test("should support filtering, lte") {
+    val expectedMap1Value: MapValue = Map("A" -> toVal(1))
+    val expectedMap2Value: MapValue = Map("A" -> toVal(2))
+    val expectedMap3Value: MapValue = Map("A" -> toVal(4))
+
+    val expectedList: ListValue = Seq(expectedMap1Value, expectedMap2Value, expectedMap3Value)
+
+    val metaMapValue: MapValue = Map("list" -> expectedList)
+
+    exec(t => {
+      val storage = t.from("mysql")
+      val table = storage.from("table1")
+      table.set("key1", metaMapValue)
+    }, commit = true)
+
+    val Seq(v) = exec(t => {
+      val storage = t.from("mysql")
+      val table = storage.from("table1")
+      t.ret(table.get("key1").filter("A", MryFilters.LesserThanOrEqual, 3))
+    }, commit = false)
+
+    val finalList = v.value.asInstanceOf[MapValue].mapValue("list").asInstanceOf[ListValue]
+
+    finalList.size should equal(2)
+
+    finalList(0) should be(MapValue(Map("A" -> toVal(1))))
+    finalList(1) should be(MapValue(Map("A" -> toVal(2))))
+  }
+
   test("should support projection on single record") {
     val expectedMapValue: MapValue = Map("mapk" -> toVal("value1"))
     exec(t => {
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
@@ -13,12 +13,13 @@ import com.wajam.nrv.service.TokenRange
 import com.wajam.nrv.utils.TimestampIdGenerator
 import com.wajam.nrv.utils.timestamp.Timestamp
 import org.mockito.Mockito._
+import org.scalatest.matchers.ShouldMatchers
 
 /**
  * Test MySQL storage
  */
 @RunWith(classOf[JUnitRunner])
-class TestMysqlStorage extends TestMysqlBase {
+class TestMysqlStorage extends TestMysqlBase with ShouldMatchers {
 
   test("should get committed record") {
     exec(t => {
@@ -39,6 +40,35 @@ class TestMysqlStorage extends TestMysqlBase {
       case _ =>
         fail("Didn't receive a map")
     }
+  }
+
+  test("should support filtering") {
+    val expectedMap1Value: MapValue = Map("A" -> toVal("1"))
+    val expectedMap2Value: MapValue = Map("A" -> toVal("2"))
+    val expectedMap3Value: MapValue = Map("A" -> toVal("2"))
+
+    val expectedList: ListValue = Seq(expectedMap1Value, expectedMap2Value, expectedMap3Value)
+
+    val metaMapValue: MapValue = Map("list" -> expectedList)
+
+    exec(t => {
+      val storage = t.from("mysql")
+      val table = storage.from("table1")
+      table.set("key1", metaMapValue)
+    }, commit = true)
+
+    val Seq(v) = exec(t => {
+      val storage = t.from("mysql")
+      val table = storage.from("table1")
+      t.ret(table.get("key1").filter("A", MryFilters.Equals, "2"))
+    }, commit = false)
+
+    val finalList = v.value.asInstanceOf[MapValue].mapValue("list").asInstanceOf[ListValue]
+
+    finalList.size should equal(2)
+
+    finalList(0) should be(MapValue(Map("A" -> toVal("2"))))
+    finalList(1) should be(MapValue(Map("A" -> toVal("2"))))
   }
 
   test("should support projection on single record") {
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
@@ -43,11 +43,11 @@ class TestMysqlStorage extends TestMysqlBase with ShouldMatchers {
   }
 
   test("should support filtering") {
-    val expectedMap1Value: MapValue = Map("A" -> toVal("1"))
-    val expectedMap2Value: MapValue = Map("A" -> toVal("2"))
-    val expectedMap3Value: MapValue = Map("A" -> toVal("2"))
+    val mapValue1: MapValue = Map("A" -> toVal("1"))
+    val mapValue2: MapValue = Map("A" -> toVal("2"))
+    val mapValue3: MapValue = Map("A" -> toVal("2"))
 
-    val expectedList: ListValue = Seq(expectedMap1Value, expectedMap2Value, expectedMap3Value)
+    val expectedList: ListValue = Seq(mapValue1, mapValue2, mapValue3)
 
     val metaMapValue: MapValue = Map("list" -> expectedList)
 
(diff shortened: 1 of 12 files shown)


mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
```diff
@@ -60,14 +82,37 @@ case class ListValue(listValue: Seq[Value]) extends Value {
...
+      } else {
+        ListValue(listValue)
+      }
+  }
+
+  override def execFiltering(context: ExecutionContext, into: Variable, key: Object, filter: MryFilters.MryFilter, value: Object) {
+
+    // Filter the child mapValue that doesn't match the filter
+    val temp = listValue filter {
+        case mapValue: MapValue =>
```
alexbergeron: Fix indent

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
```diff
@@ -60,14 +82,37 @@ case class ListValue(listValue: Seq[Value]) extends Value {
...
+        }))
+      } else {
+        ListValue(listValue)
+      }
+  }
+
+  override def execFiltering(context: ExecutionContext, into: Variable, key: Object, filter: MryFilters.MryFilter, value: Object) {
+
+    // Filter the child mapValue that doesn't match the filter
+    val temp = listValue filter {
```
alexbergeron: Use withFilter

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
```diff
@@ -60,14 +82,37 @@ case class ListValue(listValue: Seq[Value]) extends Value {
...
+        case mapValue: MapValue =>
+          mapValue.shouldFilter(key, filter, value)
+        case v: Value => true
+      }
+
+    val temp2 = temp.map {
```
alexbergeron: Chain the call with map

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
```diff
@@ -60,14 +82,37 @@ case class ListValue(listValue: Seq[Value]) extends Value {
...
+      case mapValue: MapValue =>
+        mapValue.execFiltering(context, into, key, filter, value)
         into.value
-      }))
-    } else {
-      ListValue(listValue)
+
+      // Doesn't support filtering
+      case v => v
     }
```
alexbergeron:
``` scala
val temp = listValue withFilter {
  case mapValue: MapValue =>
    mapValue.shouldFilter(key, filter, value)
  case v: Value => true
} map {
  case mapValue: MapValue => {
    mapValue.execFiltering(context, into, key, filter, value)
    into.value
  }
  case v => v
}
```

alexbergeron: Or use collect

``` scala
val temp = listValue collect {
  case mapValue: MapValue if (mapValue.shouldFilter(key, filter, value)) => {
    mapValue.execFiltering(context, into, key, filter, value)
    into.value
  }
  case v: Value if !(v.isInstanceOf[MapValue]) => v
}
```

oliviertoupin:
Added support for proxies AKA make filtering work in edge

mry-core/src/main/scala/com/wajam/mry/execution/Transaction.scala
```diff
@@ -37,6 +37,7 @@ class Transaction(blockCreator: (Block with OperationApi) => Unit = null) extend
 
   override def execReturn(context: ExecutionContext, from: Seq[Variable]) {
 
+    var i = 0
```
alexbergeron: ???

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
alexbergeron: Use withFilter!

mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
```diff
@@ -41,6 +42,64 @@ class TestMysqlStorage extends TestMysqlBase {
...
+      table.set("key1", metaMapValue)
+    }, commit = true)
+
+    val Seq(v) = exec(t => {
+      val storage = t.from("mysql")
+      val table = storage.from("table1")
+      t.ret(table.get("key1").filter("A", MryFilters.Equals, "2"))
+    }, commit = false)
+
+    val finalList = v.value.asInstanceOf[MapValue].mapValue("list").asInstanceOf[ListValue]
```
coderunner: I'm sure you can simplify this code using the fact that MapValue and ListValue are case classes.

mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
```diff
@@ -41,6 +42,64 @@ class TestMysqlStorage extends TestMysqlBase {
     }
   }
 
+  test("should support filtering") {
+    val expectedMap1Value: MapValue = Map("A" -> toVal("1"))
```
coderunner: why prefixing the values with expected?

mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
```diff
@@ -41,6 +42,64 @@ class TestMysqlStorage extends TestMysqlBase {
...
+      val storage = t.from("mysql")
+      val table = storage.from("table1")
+      t.ret(table.get("key1").filter("A", MryFilters.Equals, "2"))
+    }, commit = false)
+
+    val finalList = v.value.asInstanceOf[MapValue].mapValue("list").asInstanceOf[ListValue]
+
+    finalList.size should equal(2)
+
+    finalList(0) should be(MapValue(Map("A" -> toVal("2"))))
```
coderunner: why not reuse the map define above?

mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
```diff
@@ -41,6 +42,64 @@ class TestMysqlStorage extends TestMysqlBase {
...
+    finalList(1) should be(MapValue(Map("A" -> toVal("2"))))
+  }
+
+  test("should support filtering, lte") {
+    val expectedMap1Value: MapValue = Map("A" -> toVal(1))
```
coderunner: same comments for this test

mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
alexbergeron:
``` scala
val ListValue(finalList) = v.value.asInstanceOf[MapValue].mapValue("list")
```

It's possible to use / abuse case classes further, but I don't think it would be cleaner

``` scala
val MapValue(mapValue) = v.value
val ListValue(finalList) = mapValue(list)
```

Don't want to pollute your test with a mapValue val?

``` scala
val ListValue(finalList) = {
  val MapValue(mapValue) = v.value
  mapValue(list)
}
```

alexbergeron: (I'd definitively go for the first one :smile:)

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
```diff
@@ -60,19 +92,39 @@ case class ListValue(listValue: Seq[Value]) extends Value {
...
+
+      // Foward the filter to allow recursion
+      .map { (v) =>
+        v.execFiltering(context, into, key, filter, value)
         into.value
-      }))
-    } else {
-      ListValue(listValue)
-    }
+      }
```
alexbergeron, line 120: IMHO, this assignment to temp looks just weird... I'd go with

``` scala
val temp = listValue withFilter { v =>
  v.execPredicate(context, into, key, filter, value)
  into.value.asInstanceOf[BoolValue].boolValue
} map { v =>
  v.execFiltering(context, into, key, filter, value)
  into.value
}
```

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
```diff
@@ -18,8 +18,19 @@ trait Value extends Object with OperationSource {
   def isNull = false
 }
 
+trait NoFiltering extends Value with OperationSource {
```
alexbergeron: I don't hate the fact that this is in a well-separated trait, but, currently, most default implementations are already done in Value. I think it might be more consistent.

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -376,6 +386,7 @@ private class InternalProtobufTranslator {
       case Type.Delete => new Operation.Delete(os, variables(0), objects:_*)
       case Type.Limit => new Operation.Limit(os, variables(0), objects:_*)
       case Type.Projection => new Operation.Projection(os, variables(0), objects:_*)
+      case Type.Filter => new Operation.Filter(os, variables(0), objects(0), MryFilters.withName(objects(1).asInstanceOf[StringValue].strValue), objects(2))
```
coderunner: maybe remove the instanceOf

mry-core/src/main/scala/com/wajam/mry/execution/MryFilters.scala
```diff
@@ -0,0 +1,56 @@
+package com.wajam.mry.execution
+
+object MryFilters extends Enumeration {
```
coderunner, line 3: I think I would go with a sealed trait and case classes since each Filter has a specific behavior. Also I think there must be a way to use scala type system to avoid the throwMatchErrors for unsupported operations.

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
coderunner: Agree with Alex, if we want to go that way, we should do it for all operations. I think there is a cleanup to do, but this would require lot of work.

oliviertoupin: We can't put it in Value, since that would override all behavior (even for RecordValue and MultipleRecord) and thus disabling the proxying for those class (see OperationApi)

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
```diff
@@ -38,12 +49,33 @@ case class MapValue(mapValue: Map[String, Value]) extends Value {
...
+      case Some(v) =>
+        MryFilters.applyFilter(v, filter, value)
+      case None => true
     }
+
+    into.value = BoolValue(result)
+  }
+
+  override def execProjection(context: ExecutionContext, into: Variable, keys: Object*) {
+    into.value =
```
coderunner: why not keep the if on the same line?, I think it is idiomatic in scala

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
```diff
@@ -60,19 +92,39 @@ case class ListValue(listValue: Seq[Value]) extends Value {
...
+        ListValue(listValue)
+      }
+  }
+
+  override def execFiltering(context: ExecutionContext, into: Variable, key: Object, filter: MryFilters.MryFilter, value: Object) {
+
+    val temp = listValue
+
+      // Filter the child mapValue that doesn't match the filter
+      .withFilter { (v) =>
```
coderunner: not sure I like to formating I think it is confusing

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
```diff
@@ -60,19 +92,39 @@ case class ListValue(listValue: Seq[Value]) extends Value {
...
+        v.execPredicate(context, into, key, filter, value)
```
coderunner, line 117: is execPredicate defined for all values?

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
coderunner, line 120: don't like temp variable name...either we name it resultingList (or something else) or we remove it

alexbergeron, line 120: I mostly concentrated on formatting, not really naming, but I could suggest

``` scala
val finalList = listValue withFilter { v =>
  v.execPredicate(context, into, key, filter, value)
  val BoolValue(filtered) = into.value
  filtered
} map { v =>
  v.execFiltering(context, into, key, filter, value)
  into.value
}
```

(I also removed to asInstanceOf)

mry-core/src/test/scala/com/wajam/mry/execution/TestOperations.scala
```diff
@@ -0,0 +1,278 @@
...
+    val listValue: ListValue = ListValue(Seq(Map("key1" -> toVal("value1"), "key2" -> toVal("value2"))))
+    val operationSource: OperationSource = listValue
+
+    val v = new Variable(null, 0)
+
+    operationSource.execProjection(null, v, StringValue("key2"))
+    v.value should be(ListValue(Seq(Map("key2" -> toVal("value2")))))
+  }
+
+  def sampleFilterList(): ListValue = {
```
coderunner: I would put util function at the end of the file to have all the tests together at the top.

mry-core/src/test/scala/com/wajam/mry/execution/TestOperations.scala
```diff
@@ -0,0 +1,278 @@
...
+    val expectedList: ListValue = Seq(expectedMap1Value, expectedMap2Value, expectedMap3Value, expectedMap4Value)
+
+    expectedList
+  }
+
+  def assertFilterResult(list: ListValue, count: Int) {
+
+    list.listValue.size should equal(count)
+
+    val target = MapValue(Map("A" -> toVal("2")))
```
coderunner: I don't like the fact that the Map are defined twice...once in the setup and once in the assert...should really be the same reference so if we change either the test still run.

mry-core/src/test/scala/com/wajam/mry/execution/TestOperations.scala
```diff
@@ -0,0 +1,278 @@
...
+
+import org.scalatest.FunSuite
+import com.wajam.mry.execution.Implicits._
+import org.scalatest.matchers.ShouldMatchers
+
+/**
+ * Test non-database dependent Mry Operations
+ */
+class TestOperations extends FunSuite with ShouldMatchers {
+
```
coderunner, line 11: Lots of tests, good coverage, I like it!

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
coderunner: Not in Value but in MapValue and ListValue like execProjection.

oliviertoupin:
Everything is fixed expect for MryFilters as classes. The implementation I have in my head would result in more code, so let's talk tomorrow about what you had in mind.

mry-core/src/main/scala/com/wajam/mry/execution/Value.scala
```diff
@@ -18,8 +18,26 @@ trait Value extends Object with OperationSource {
...
+ * It allow Mry proxying (see OperationSource (getProxiedSource.execPredicate)) to work.
+ * Without it concrete class would throw a runtime exception.
+ *
+ * It avoid duplication and it allow smooth transition from no support to full predicate support, in the concrete class.
+ *
+ * TODO: Implement predicate on all values.
+ */
+trait DefaultPredicate extends Value with OperationSource {
+
+  override def execPredicate(context: ExecutionContext, into: Variable, key: Object, filter: MryFilters.MryFilter, value: Object) = {
```
coderunner, line 33: indent

coderunner:
+1

PR #118, opened 2013-05-06
Update dependencies for Scala 2.10

alexbergeron:


alexbergeron:
Rebasing

mry-core/src/main/scala/com/wajam/mry/ConsistentDatabase.scala
```diff
@@ -60,7 +60,7 @@ class ConsistentDatabase[T <: ConsistentStorage](serviceName: String = "database
 
     readTransactionsInitTimer.time {
       new Iterator[Message] with Closable {
-        val itr = storage.readTransactions(from, to, ranges)
+        val itr = storage.readTransactions(from, to.asInstanceOf[Timestamp], ranges)
```
coderunner: really need the asInstanceOf???

alexbergeron: The problem was met in Scala 2.10. It's probably hiding another problem, will investigate then.

Since the main point of theses PR are to update the dependencies, not to update to 2.10, I will revert theses changes.

coderunner:
+1

PR #122, opened 2013-05-21
GC  lookup last consistent timestamp of the current range instead of all ranges

mlaflamm:
Also added a meter for GC error...


mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -687,6 +687,8 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
       "gc-collected-record", table.uniqueName, "records", TimeUnit.SECONDS))
     lazy private val extraVersionsLoadedMeter = new Meter(metrics.metricsRegistry.newMeter(GarbageCollector.getClass,
       "gc-extra-record-loaded", table.uniqueName, "extra-record-loaded", TimeUnit.SECONDS))
+    lazy private val collectError = new Meter(metrics.metricsRegistry.newMeter(GarbageCollector.getClass,
```
alexbergeron, line 690: Wondering why you are using Metrics directly instead of metrics-scala

mlaflamm, line 690: This has been introduced in a previous refactoring to keep the original scope (i.e. package + class) used prior that refactoring.

alexbergeron:
+1

PR #123, opened 2013-05-27
[Mry] Fix empty resultset being null

oliviertoupin:


mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -197,8 +197,9 @@ private class InternalProtobufTranslator {
         for (v <- values) {
           val value = encodePValue(v)
           pTransport.addResponseHeapIds(addToHeap(value))
-          pTransport.setType(Type.Values)
         }
+
+        pTransport.setType(Type.Values)
```
coderunner: Set the type as the first statement in the case would be slighty better

mry-core/src/test/scala/com/wajam/mry/api/protobuf/TestProtobufTranslator.scala
```diff
@@ -172,7 +172,22 @@ class TestProtobufTranslator extends FunSuite with ShouldMatchers {
...
+    val transport = new Transport(None, Some(results))
+
+    val bytes = translator.encodeAll(transport)
+    val transport2 = translator.decodeAll(bytes)
+
+    val results2 = transport2.values.get
+
+    transport2.transaction.isDefined should be(false)
 
+    results equalsValue results2
```
coderunner: don't need the equalsValue that will implicitely convert to ListValue

coderunner:
+1

PR #125, opened 2013-06-04
Fix replication slow queries

mlaflamm:
Use a timestamp upper bound of 24 hours when fetching table summary records.


coderunner:
+1

PR #128, opened 2013-07-19
New default Database current consistent timestamps hook

mlaflamm:
Now allows MysqlStorage percolation and GC on all data by default. Doesn't need to be hooked with ConsistencyMasterSlave anymore.

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
@@ -26,7 +26,7 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
   private[mysql] var transactionMetrics: MysqlTransaction.Metrics = null
   private var tablesMutationsCount = Map[Table, AtomicInteger]()
   val valueSerializer = new ProtobufTranslator
-  private var currentConsistentTimestamps: (TokenRange) => Timestamp = (_) => Long.MinValue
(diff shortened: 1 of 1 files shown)


mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -26,7 +26,7 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
   private[mysql] var transactionMetrics: MysqlTransaction.Metrics = null
   private var tablesMutationsCount = Map[Table, AtomicInteger]()
   val valueSerializer = new ProtobufTranslator
-  private var currentConsistentTimestamps: (TokenRange) => Timestamp = (_) => Long.MinValue
+  private var currentConsistentTimestamps: (TokenRange) => Timestamp = (_) => Long.MaxValue
```
oliviertoupin, line 29: I'm not sure to understand the link with this change and the description of the changes.

mlaflamm, line 29: MRY doesn't percolate or GC beyond the current consistent timestamp. Having MinValue by default means that no data is ever percolated or GCed unless the default is modified. The default is modified only when consistency and transaction log is enabled. 

With this change percolation and GC will be possible by default unless used with ConsistencyMasterSlave.

alexbergeron:
I trust you on that, +1

PR #129, opened 2013-07-24
TableContinuousFeeder is now a mixable trait

mlaflamm:
TableContinuousFeeder has been rewritten and split in two: a TableContinuousFeeder mixable trait and TableAllLatestFeeder which load the data. 

Use the two together to have the same functionality as before: <code>new TableAllLatestFeeder(...) with TableContinuousFeeder</code>

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/storage/mysql/RecordDataFeeder.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/storage/mysql/RecordDataFeeder.scala
@@ -21,9 +21,11 @@ trait ResumableRecordDataFeeder extends RecordDataFeeder {
 
   def loadRecords(range: TokenRange, fromRecord: Option[DataRecord]): Iterable[DataRecord]
 
+  def toContextData(data: Map[String, Any]): Map[String, Any] = data
+
   def ack(data: Map[String, Any]) {
     // Update context with the latest acknowledged record data
-    context.data = data.map(entry => entry match {
+    context.data = toContextData(data).map(entry => entry match {
       case (k, v: String) => (k, v)
       case (k, v: Seq[Any]) => (k, v.map(_.toString))
       case (k, v) => (k, v.toString)
--- a/mry-core/src/main/scala/com/wajam/mry/storage/mysql/RecordDataFeeder.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/storage/mysql/RecordDataFeeder.scala
@@ -1,0 +1,33 @@
+package com.wajam.mry.storage.mysql
+
+import com.wajam.nrv.service.TokenRange
+import com.wajam.spnl.feeder.Feeder
+import com.wajam.spnl.TaskContext
+
+trait RecordDataFeeder extends Feeder {
+
+  type DataRecord
+
+  def toRecord(data: Map[String, Any]): Option[DataRecord]
+
+  def toData(record: DataRecord): Map[String, Any]
+}
+
+trait ResumableRecordDataFeeder extends RecordDataFeeder {
+
+  def context: TaskContext
+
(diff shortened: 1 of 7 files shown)


mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableContinuousFeeder.scala
```diff
@@ -1,119 +1,64 @@
...
   def loadMore() = {
-
-    var transaction: MysqlTransaction = null
     try {
-      val fromRecord = lastRecord
-      currentRange = lastRecord match {
-        case Some(record) => tokenRanges.find(_.contains(record.token))
+      val loadRange = lastRecord match {
+        case Some(record) => {
+          println("1) lastRecord=%s, curRange=%s, newRange=%s".format(lastRecord, currentRange,
```
appaquet: Yep...

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableContinuousFeeder.scala
```diff
@@ -1,119 +1,64 @@
...
+            tokenRanges.find(_.contains(token(record))).getOrElse(tokenRanges.head)))
+          tokenRanges.find(_.contains(token(record))).getOrElse(tokenRanges.head)
+        }
         case None => {
           currentRange match {
-            case Some(range) => range.nextRange(tokenRanges)
-            case None => Some(tokenRanges.head)
+            case Some(range) => {
+              println("2) lastRecord=%s, curRange=%s, newRange=%s".format(lastRecord, currentRange,
```
appaquet: Yep again...

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableContinuousFeeder.scala
alexbergeron: println???

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableAllLatestFeeder.scala
```diff
@@ -0,0 +1,69 @@
...
+      transaction.getAllLatest(table, loadLimit, range, fromRecord).toList
+    } finally {
+      transaction.commit()
+    }
+  }
+
+  def token(record: Record) = record.token
+
+  def toRecord(data: Map[String, Any]) = {
+    if (data.contains(Keys) && data.contains(Token) && data.contains(Timestamp))
```
coderunner, line 31: toRecord does not populate the Value?

mry-core/src/main/scala/com/wajam/mry/storage/mysql/RecordDataFeeder.scala
```diff
@@ -0,0 +1,33 @@
...
+
+trait ResumableRecordDataFeeder extends RecordDataFeeder {
+
+  def context: TaskContext
+
+  def token(record: DataRecord): Long
+
+  def loadRecords(range: TokenRange, fromRecord: Option[DataRecord]): Iterable[DataRecord]
+
+  def ack(data: Map[String, Any]) {
```
coderunner, line 26: Shouldn't ack() call an abstract method that takes the data and return the Map of what to save? So that child class do not have to hack ack() :)

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableAllLatestFeeder.scala
mlaflamm, line 31: Good catch! I extracted this from the existing code which wasn't populating the Value. I added it for completeness.

coderunner:
+1

PR #130, opened 2013-07-29
Tombstone record deleter

mlaflamm:


mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
```diff
@@ -580,6 +580,105 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
...
+      case None => ("tk >= %d".format(range.start), Seq())
+    }
+    val maxTimestamp: Timestamp = storage.getCurrentConsistentTimestamp(Seq(range)).value - minTombstoneAge
+
+    /*
+     * Generated SQL looks like:
+     *
+     *    SELECT tk, ts, k1
+     *    FROM `table1_data`
+     *    WHERE (((tk > 4025886270)) OR ((tk = 4025886270) AND (k1 > 'key15')))
```
coderunner, line 610: Why the predicate (tk = 4025886270) AND (k1 > 'key15')? To start from a previous position? Then shouldn't we have order by clauses?

mlaflamm, line 610: Assuming we have 10000 records for a given token but the select is limited to 100, we want to resume from the previously returned record. This is the exact same behavior than `getAllLatests(...)` which also uses `genWhereHigherEqualTuple(...)`

mlaflamm, line 610: Just reread the question more carefully. There is only one index for this table but you are probably right, I should explicitly add an ORDER BY clause.

coderunner:
+1

PR #131, opened 2013-08-12
Extracted entity model from demo

alexbergeron:
@mlaflamm @jnblanchet


mry-core/src/main/scala/com/wajam/mry/api/entities/MryEntity.scala
```diff
@@ -0,0 +1,339 @@
+package com.wajam.mry.api.entities
```
mlaflamm: Why in api and why plural? I suggest `com.wajam.mry.entity`

mry-core/src/main/scala/com/wajam/mry/api/entities/MryEntity.scala
```diff
@@ -0,0 +1,339 @@
...
+  override def init() {
+    if (isDefined)
+      super.init()
+  }
+
+  override def toString: String = name + "=" + (if (isDefined) super.toString else "undefined")
+}
+
+
+abstract class MryEntity extends FieldCollection {
```
mlaflamm: I suggest simply Entity.

mlaflamm:
Where are the unit tests? :smile:

mry-core/src/main/scala/com/wajam/mry/api/entities/MryEntity.scala
alexbergeron: Was looking to put it in an existing package (well, creating its own subpackage)

 But I guess it would be nice to also separate theses in multiple files for similar concepts

mry-core/src/main/scala/com/wajam/mry/api/entities/MryEntity.scala
alexbergeron: I put MRY in it because Entity was kind of generic, but both makes sense

alexbergeron:
@mlaflamm: Unit tests are in @appaquet's head

mry-core/src/main/scala/com/wajam/mry/api/entities/MryEntity.scala
mlaflamm: Well many things in MRY have generic names (e.g. Storage, Database, Transport, Block, Result, etc) and this is fine. We should be consistent and omit it or add it everywhere. I am personally not fond of this kind of prefix, the package is sufficient in most cases.

mlaflamm:
Sorry @alex-wajam, you are not @appaquet, so no free ride for you...

alexbergeron:
But I want to extract the Unit test from @appaquet's head!

alexbergeron:
I'm removing the DateTime field support, I'd prefer that the applications using it define how dates are managed.

appaquet:
LOL!

mry-core/src/main/scala/com/wajam/mry/entities/Entity.scala
```diff
@@ -0,0 +1,239 @@
+package com.wajam.mry.entities
```
mlaflamm: Again why plurals? I don't think we use plurals in the frameworks.

alexbergeron:
Will rework this to make more immutable...

appaquet:
I don't like the fact that the Models/Entities are going to be directly in MRY. It's kind of like if Cassandra or Hbase was putting in their core code Java entity model for their database. MRY should remain language/design agnostic, no? @coderunner @mlaflamm, what do you think? I'd propose to create a mry-scala-entity repo?

alexbergeron:
com.wajam.commons.mry???
(Subproject of wajam-commons repo that exists in my dreams and solve all problems we have in this world)

appaquet:
I'd call the repo "scala-commons" and then com.wajam.commons.mry so that we don't confuse PHPers and C++ers :)

coderunner:
Depends on the granularity we want, but I guess we would like to OS the entities as well. Usually commons repo gets filled with lots of random stuff. Maybe a public repo mry-entities or mry-utils?

appaquet:
Yeah, I had said mry-scala-entities because mry could be used in another language, but mry-entities or mry-utils makes sense too

alexbergeron:
com.wajam.mryentities???

mry-core/src/main/scala/com/wajam/mry/entity/Entity.scala
```diff
@@ -0,0 +1,215 @@
...
+trait MryConvertible {
+
+  def toMry: Value
+
+  def fromMry(value: Value): this.type
+}
+
+trait FieldLike extends MryConvertible {
+
+  def validate(): Option[String] = {
```
appaquet: I'd still keep a method that throws. Personally, I prefer throwing exceptions in code where I expect a model to be complete rather than having to case match the validate() method everytime and then throw. Why not a checkValid method that returns a scala Try and then validate() that throws?

mry-core/src/main/scala/com/wajam/mry/entity/Entity.scala
```diff
@@ -0,0 +1,215 @@
...
+  def clear() {
+    _list = Seq()
+  }
+
+  def apply(pos: Int) = _list(pos)
+
+  def fromMry(value: Value): this.type = {
+    value match {
+      case s: ListValue => this.set(s.listValue.map(m.fromMry))
+      case _ => throw new ClassCastException("Cannot load field %s with value %s".format(name, value))
```
appaquet: Why a format here and string interpolation elsewhere?

mry-core/src/main/scala/com/wajam/mry/entity/FieldTypeHandler.scala
```diff
@@ -0,0 +1,90 @@
...
+import com.wajam.mry.execution._
+
+trait FieldTypeHandler[T] {
+  def getDefault(v: Field[T]): Option[T] = None
+
+
+  def toMry(v: T): Value
+
+  def fromMry(value: Value): T
+}
```
appaquet: Probably didn't run a format on that :P

mry-core/src/main/scala/com/wajam/mry/entity/Entity.scala
alexbergeron: Because I missed it while reworking the code

alexbergeron: Actually, mry still "supports" 2.9.1, should use formatiing...

mry-entity/src/main/scala/com/wajam/mry/entity/Entity.scala
```diff
@@ -0,0 +1,216 @@
...
+trait MryConvertible {
+
+  def toMry: Value
+
+  def fromMry(value: Value): this.type
+}
+
+trait FieldLike extends MryConvertible {
+
+  def validate(): Option[String] = {
```
appaquet, line 16: My previous comment on validate() is hidden, what do you think about my thought? @alex-wajam @coderunner ?

alexbergeron, line 16: I'm honestly kinda confused on it. I tried to make it throw as late as possible, so we could easily do some further refactoring. In my ideal world a model that was not set completely would fail at compile time, not run time.

I'd vote for throwing as late as possible

appaquet:
Todo: https://gist.github.com/coderunner/f65320643fbfc9f3617a ?? :)

alexbergeron:
@appaquet : I've planned some time for it this sprint

alexbergeron:
_discretely closes 2 year old PR on seemingly no longer maintained library_

appaquet:
Nooooooooooo!

PR #134, opened 2013-09-04
Fix warnings for MRY

alexbergeron:


mry-core/src/main/scala/com/wajam/mry/execution/Implicits.scala
```diff
@@ -69,8 +71,8 @@ object Implicits {
       case v: Long => v
       case v: Boolean => v
       case v: Double => v
-      case v: Map[String, Any] => v
-      case v: Seq[Any] => v
+      case v: Map[_, _] => v.asInstanceOf[Map[String, Any]]
+      case v: Seq[_] => v
```
coderunner, line 75: Why do you asInstanceOf the map case, but not the Seq case?

coderunner:
+1

project/Project.scala
```diff
@@ -39,7 +39,8 @@ object MryBuild extends Build {
     publishMavenStyle := true,
     organization := "com.wajam",
     version := "0.1-SNAPSHOT",
-    scalaVersion := "2.10.2"
+    scalaVersion := "2.10.2",
+    scalacOptions ++= Seq("-deprecation", "-unchecked", "-feature")
```
oliviertoupin, line 43: Could that hide any future warning?

alexbergeron, line 43: Exactly the opposite: replaces the "There are 3 features warnings" by theses 3 features warnings

mry-core/src/main/scala/com/wajam/mry/execution/Implicits.scala
alexbergeron, line 75: Seq[_](Since Seq[Any] is accepted) achieves to call the specific implicit, while Map (Which hides the string key) doesn't

PR #135, opened 2013-09-06
Use Scala futures

alexbergeron:


mry-core/src/test/scala/com/wajam/mry/TestDatabase.scala
```diff
@@ -52,12 +62,13 @@ class TestDatabase extends FunSuite with BeforeAndAfterAll {
         b.from("memory").set(key, "value%s".format(key))
       })
 
+
       db.execute(b => {
         val context = b.from("context")
         b.returns(b.from("memory").get(key), context.get("tokens"), context.get("local_node"))
-      }, p.complete(_, _))
+      }, complete(p))
```
coderunner: Shouldn't the test use the execute methods that returns a Future?

alexbergeron: +1

alexbergeron:
Updated

coderunner:
+1

PR #136, opened 2013-09-06
Fix TableAllLatestFeeder when deleted records range > loadLimit

mlaflamm:
TableAllLatestFeeder wasn't iterating over the complete table when deleted records range was > loadLimit.

### DIFF
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
@@ -1104,15 +1104,17 @@ class TestMysqlStorage extends TestMysqlBase with ShouldMatchers {
       val t1 = t.from("mysql").from("table1")
       deletedKeys.foreach(tup => t1.delete(tup._2))
       createdKeys.foreach(tup => t1.set(tup._2, Map(tup._2 -> tup._2)))
-    }, commit = true)
+    }, commit = true, onTimestamp = createTimestamp(0))
 
     var recordsExcludeDeleted = mysqlStorage.createStorageTransaction(context).getAllLatest(table1, 50).toList
-    recordsExcludeDeleted.size should be(createdKeys.size)
 
     var recordsIncludeDeleted = mysqlStorage.createStorageTransaction(context).getAllLatest(table1, 50, includeDeleted = true).toList
-    recordsIncludeDeleted.size should be(keys.size)
 
     val recordsDeleted = recordsIncludeDeleted.filter(_.value == NullValue)
+    println("%d, %d, %d".format(recordsExcludeDeleted.size, recordsIncludeDeleted.size, recordsDeleted.size))
+
+    recordsExcludeDeleted.size should be(createdKeys.size)
+    recordsIncludeDeleted.size should be(keys.size)
     recordsDeleted.size should be(deletedKeys.size)
   }
 
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
@@ -1112,7 +1112,7 @@ class TestMysqlStorage extends TestMysqlBase with ShouldMatchers {
     var recordsIncludeDeleted = mysqlStorage.createStorageTransaction(context).getAllLatest(table1, 50, includeDeleted = true).toList
     recordsIncludeDeleted.size should be(keys.size)
 
-    val recordsDeleted = recordsIncludeDeleted.filter(_.value == NullValue)
+    val recordsDeleted = recordsIncludeDeleted.filter(_.value.isNull)
     recordsDeleted.size should be(deletedKeys.size)
   }
 
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
@@ -1107,14 +1107,12 @@ class TestMysqlStorage extends TestMysqlBase with ShouldMatchers {
     }, commit = true, onTimestamp = createTimestamp(0))
 
     var recordsExcludeDeleted = mysqlStorage.createStorageTransaction(context).getAllLatest(table1, 50).toList
+    recordsExcludeDeleted.size should be(createdKeys.size)
 
     var recordsIncludeDeleted = mysqlStorage.createStorageTransaction(context).getAllLatest(table1, 50, includeDeleted = true).toList
+    recordsIncludeDeleted.size should be(keys.size)
 
     val recordsDeleted = recordsIncludeDeleted.filter(_.value == NullValue)
-    println("%d, %d, %d".format(recordsExcludeDeleted.size, recordsIncludeDeleted.size, recordsDeleted.size))
-
-    recordsExcludeDeleted.size should be(createdKeys.size)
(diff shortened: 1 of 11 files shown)


mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableAllLatestFeeder.scala
```diff
@@ -152,15 +144,13 @@ class TestTableAllLatestFeeder extends TestMysqlBase {
     val context2 = new TaskContext()
     context2.updateFromJson(feeder1.context.toJson)
     feeder2.init(context2)
-    var records2 = Iterator.continually({
-      feeder2.next()
-    }).take(10).flatten.toList
+    var records2 = feeder2.take(10).flatten.toList
```
coderunner: var?

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableAllLatestFeeder.scala
```diff
@@ -16,15 +16,39 @@ abstract class TableAllLatestFeeder(val name: String, storage: MysqlStorage, tab
...
     } finally {
       transaction.commit()
     }
   }
 
+  @tailrec
+  private def loadFirstNonDeletedRecords(range: TokenRange, startAfterRecord: Option[Record])(
+                                         implicit transaction: MysqlTransaction): Option[Record] = {
+    val records = transaction.getAllLatest(table, loadLimit, range, startAfterRecord, includeDeleted = true)
+    val filteredRecords = startAfterRecord match {
```
coderunner: extract method?

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableAllLatestFeeder.scala
```diff
@@ -16,15 +16,39 @@ abstract class TableAllLatestFeeder(val name: String, storage: MysqlStorage, tab
...
+      case Some(startRecord) => records.filter(_ != startRecord).toList
+      case None => records.toList
+    }
+    filteredRecords.collectFirst{case record if record.value != NullValue => record} match {
+      case record@Some(_) => record
+      case None if filteredRecords.isEmpty => None
```
coderunner: I would put the if before and collect in the else so that the logic is clearer.

mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestMysqlStorage.scala
```diff
@@ -1090,6 +1090,34 @@ class TestMysqlStorage extends TestMysqlBase with ShouldMatchers {
...
+      val t1 = t.from("mysql").from("table1")
+      deletedKeys.foreach(tup => t1.delete(tup._2))
+      createdKeys.foreach(tup => t1.set(tup._2, Map(tup._2 -> tup._2)))
+    }, commit = true, onTimestamp = createTimestamp(0))
+
+    var recordsExcludeDeleted = mysqlStorage.createStorageTransaction(context).getAllLatest(table1, 50).toList
+
+    var recordsIncludeDeleted = mysqlStorage.createStorageTransaction(context).getAllLatest(table1, 50, includeDeleted = true).toList
+
+    val recordsDeleted = recordsIncludeDeleted.filter(_.value == NullValue)
```
alexbergeron: `_.value.isNull`

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableAllLatestFeeder.scala
```diff
@@ -16,15 +16,39 @@ abstract class TableAllLatestFeeder(val name: String, storage: MysqlStorage, tab
...
 
+  @tailrec
+  private def loadFirstNonDeletedRecords(range: TokenRange, startAfterRecord: Option[Record])(
+                                         implicit transaction: MysqlTransaction): Option[Record] = {
+    val records = transaction.getAllLatest(table, loadLimit, range, startAfterRecord, includeDeleted = true)
+    val filteredRecords = startAfterRecord match {
+      case Some(startRecord) => records.filter(_ != startRecord).toList
+      case None => records.toList
+    }
+    filteredRecords.collectFirst{case record if record.value != NullValue => record} match {
```
alexbergeron: `if !record.value.isNull`

mlaflamm:
Updated

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableAllLatestFeeder.scala
```diff
@@ -16,15 +16,41 @@ abstract class TableAllLatestFeeder(val name: String, storage: MysqlStorage, tab
...
+    implicit val transaction = storage.createStorageTransaction
     try {
-      transaction.getAllLatest(table, loadLimit, range, fromRecord).toList
+      val records = filterStartRecord(
+        transaction.getAllLatest(table, loadLimit, range, startAfterRecord).toIterable, startAfterRecord).toList
+      if (records.isEmpty) {
+        // No records found! We don't know if we have reach the end of the table or if the number of consecutive
+        // deleted records is larger than the loadLimit. Fallback to a slower but deterministic method.
+        // Load all following records including deleted records until we reach a non deleted record or the end of the table
+        loadFirstNonDeletedRecords(range, startAfterRecord)
```
appaquet: Reason other than laziness to use implicit transaction?

mlaflamm: Not really. I agree this is not necessary.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableAllLatestFeeder.scala
```diff
@@ -16,15 +16,41 @@ abstract class TableAllLatestFeeder(val name: String, storage: MysqlStorage, tab
...
+      }
     } finally {
       transaction.commit()
     }
   }
 
+  @tailrec
+  private def loadFirstNonDeletedRecords(range: TokenRange, startAfterRecord: Option[Record],
+                                         transaction: MysqlTransaction): Option[Record] = {
+    val records = filterStartRecord(
```
alexbergeron, line 40: Personally, I use the opening parenthesis of a second (or third or etc) param list on a new line, not the first parameter. 
That way, it's explicit that it's another parameter list.

This is mostly minor and mostly a matter of taste, but I wonder if we should try to make this uniform or if we just leave it as it is since it's not a major point.

mlaflamm, line 40: I usually do the same but in this case the first parameter was a call to another method, `transaction.getAllLatest`,  that was still too long with all its parameters. I prefer to not break in the middle of a second level method call as this is not clear on the second line if the params are for the 1st or 2nd level method...

alexbergeron, line 40: I don't think we are talking about the same line :wink:

I was typing this while you updated your PR, I'm talking about a method signature with a second parameter list (like for currying or implicit parameters)

alexbergeron, line 40: You just changed it to a single-parameter list. We've avoided the discussion... for now!

mlaflamm, line 40: Just took a look at the previous code version and I totally agree with you about parenthesis placement with currying...

alexbergeron:
:+1:

PR #139, opened 2013-10-03
Refactored Feeders using SPNL's TaskData

clementgarnier:


mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableAllLatestFeeder.scala
```diff
@@ -32,7 +32,7 @@ class TestTableAllLatestFeeder extends TestMysqlBase {
     // 91 = 100 - (5 loadMore + 4 end of records)
     records.size should be(91)
 
-    val strKeys = records.map(_(Keys).asInstanceOf[Seq[String]](0))
+    val strKeys = records.map(_.fields(Keys).asInstanceOf[Seq[String]](0))
```
alexbergeron: `(0)` could be replaced by `.head`

alexbergeron:
+1, but I would love @mlaflamm to give it a quick check in case I might miss some bad usage of TaskData

mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableAllLatestFeeder.scala
mlaflamm: Agree

mlaflamm:
+1

PR #140, opened 2013-10-03
Performance improvement in ProtobufTranslator

oliviertoupin:
- Using prepend and a reverse list for encoding
- Same for decoding but use it's own for clarity

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
@@ -68,7 +68,8 @@ private class InternalProtobufTranslator {
   private val pb2obj = new collection.mutable.HashMap[Int, AnyRef] // Encoded HeapId to decoded instance mapping
   private val obj2pb = new collection.mutable.HashMap[AnyRef, Int] // Live instance to encoded HeapId mapping
 
-  private var tempHeap = Seq[AnyRef]()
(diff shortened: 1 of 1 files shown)


mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -114,7 +115,7 @@ private class InternalProtobufTranslator {
 
     val instanceHeapId = currentHeapId
 
-    tempHeap = tempHeap :+ mryData
+    reverseEncodingHeap = mryData +: reverseEncodingHeap
```
alexbergeron: Use `::`, it's the official list prepend method (and kinda faster from what I've seen in Scala code)

mry-core/src/main/scala/com/wajam/mry/api/protobuf/ProtobufTranslator.scala
```diff
@@ -148,15 +151,17 @@ private class InternalProtobufTranslator {
           case pMryData if pMryData.hasValue => pMryData.getValue
         }
 
-        tempHeap = tempHeap :+ value
+        reverseDecodingHeap = value +: reverseDecodingHeap
```
alexbergeron: `::`

coderunner:
Looks good to me +1

alexbergeron:
I still insist on using `::`

coderunner:
+::

PR #142, opened 2013-10-07
Optimize synthetic deleted generated when replicating from storage

mlaflamm:
Do not generate child record delete when parent is deleted in the same transaction.

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
@@ -1,7 +1,6 @@
 package com.wajam.mry.storage.mysql
 
 import com.mchange.v2.c3p0.ComboPooledDataSource
-import com.wajam.nrv.Logging
 import java.sql.{PreparedStatement, ResultSet, SQLException, Connection}
 import com.wajam.mry.execution._
 import com.wajam.mry.api.protobuf.ProtobufTranslator
@@ -13,7 +12,7 @@ import java.util.concurrent.{TimeUnit, ScheduledThreadPoolExecutor}
 import com.wajam.nrv.service.TokenRange
 import com.yammer.metrics.core.Gauge
 import annotation.tailrec
-import com.wajam.commons.Closable
+import com.wajam.commons.{Logging, Closable}
 import com.wajam.nrv.utils.timestamp.Timestamp
 
 /**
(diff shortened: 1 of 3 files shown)


mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
```diff
@@ -393,14 +392,21 @@ class MysqlStorage(config: MysqlStorageConfiguration, garbageCollection: Boolean
      */
     def applyTo(transaction: Transaction) {
       import com.wajam.mry.execution.Implicits._
+      import java.util
 
       // Set or delete each record
       val storage = transaction.from("mysql")
+      val deleted: util.NavigableSet[CompositeKey[(String, String)]] = new util.TreeSet()
```
alexbergeron: Why java TreeSet?

mlaflamm: http://docs.oracle.com/javase/7/docs/api/java/util/TreeSet.html#subSet(E, boolean, E, boolean)

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
```diff
@@ -952,6 +952,25 @@ class MutationRecord {
   }
 }
 
+case class CompositeKey[T](keys: T*)(implicit ord: math.Ordering[T]) extends Comparable[CompositeKey[T]] {
```
coderunner, line 955: Should we put this in another file? MysqlTransaction is already about 1k lines.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlStorage.scala
alexbergeron: Using a Scala sorted set

``` scala
deleted.dropWhile(_ < deleteKey.head).takeWhile(_ <= deleteKey.parent)
```

mlaflamm:
I found a flaw in the solution when reflecting more about the usage of java TreeSet as raised by @alex-wajam. I will come back soon with something that actually works!

alexbergeron:
Always my pleasure to make sure Java collections are used less and less

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
mlaflamm, line 955: I think we should move much more than this class if we want to go that way. For consistency sake, either we keep it in this file or also move AccessKey, AccessPath, Record, MutationRecord, RecordIterator, VersionRecord, etc...

coderunner:
+1

alexbergeron:
Makes a lot more sense. 

+1

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
alexbergeron, line 955: +1

PR #143, opened 2013-10-08
Added id field in TaskData

clementgarnier:


mry-core/src/main/scala/com/wajam/mry/storage/mysql/RecordDataFeeder.scala
```diff
@@ -22,9 +22,9 @@ trait ResumableRecordDataFeeder extends RecordDataFeeder {
 
   def loadRecords(range: TokenRange, startAfterRecord: Option[DataRecord]): Iterable[DataRecord]
 
-  def toContextData(data: TaskData): ContextData = data.values + ("token" -> data.token)
+  def toContextData(data: TaskData): ContextData = data.values + ("token" -> data.token) + ("id" -> data.id)
 
-  def fromContextData(data: ContextData): TaskData = TaskData(data("token").toString.toLong, values = data - "token")
+  def fromContextData(data: ContextData): TaskData = TaskData(data("token").toString.toLong, id = data("id").toString.toLong, values = data - "token")
```
mlaflamm: This will likely fail the first time we load the data from ZK after upgrading as id is not persisted by the previous version.

mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableContinuousFeeder.scala
```diff
@@ -39,7 +39,7 @@ class TestTableContinuousFeeder extends FunSuite {
 
     def toRecord(data: TaskData) = Some(data.token)
 
-    def fromRecord(record: Long) = TaskData(token = record)
+    def fromRecord(record: Long) = TaskData(token = record, id = 0)
```
mlaflamm: Hard coding the `id` works but this is not ideal. The `id` should be unique. I suggest to change the `DataRecord` type to `(Long, Long)` and to increment the id in `loadRecords`

mry-core/src/main/scala/com/wajam/mry/storage/mysql/RecordDataFeeder.scala
mlaflamm: A solution is to change the signature to `Option[TaskData]` and returns `None` if either `"token"` or `"id"` field is missing. This will also take care of the ugly `if` in `TableContinuousFeeder.init(...)`

oliviertoupin:
I have some trouble understanding what this PR is meant to be.

clementgarnier:
@oliviertoupin: It was supposed to sync MRY with the updates we were making in SPNL, but we changed some design in SPNL in the mean time, so I'm now reverting some of this stuff.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/RecordDataFeeder.scala
```diff
@@ -2,16 +2,16 @@ package com.wajam.mry.storage.mysql
...
 
 trait RecordDataFeeder extends Feeder {
 
   type DataRecord
 
-  def toRecord(data: TaskData): Option[DataRecord]
+  def toRecord(data: FeederData): Option[DataRecord]
 
-  def fromRecord(record: DataRecord): TaskData
+  def fromRecord(record: DataRecord): Map[String, Any]
```
mlaflamm: Return type should be `FeederData`

mry-core/src/main/scala/com/wajam/mry/storage/mysql/RecordDataFeeder.scala
```diff
@@ -22,11 +22,9 @@ trait ResumableRecordDataFeeder extends RecordDataFeeder {
 
   def loadRecords(range: TokenRange, startAfterRecord: Option[DataRecord]): Iterable[DataRecord]
 
-  def toContextData(data: TaskData): ContextData = data.values + ("token" -> data.token)
+  def toContextData(data: FeederData): Map[String, Any] = data
 
-  def fromContextData(data: ContextData): TaskData = TaskData(data("token").toString.toLong, values = data - "token")
-
-  def ack(data: TaskData) {
+  def ack(data: Map[String, Any]) {
```
mlaflamm: Should also be `FeederData` here

mry-core/src/main/scala/com/wajam/mry/storage/mysql/RecordDataFeeder.scala
```diff
@@ -22,11 +22,9 @@ trait ResumableRecordDataFeeder extends RecordDataFeeder {
 
   def loadRecords(range: TokenRange, startAfterRecord: Option[DataRecord]): Iterable[DataRecord]
 
-  def toContextData(data: TaskData): ContextData = data.values + ("token" -> data.token)
+  def toContextData(data: FeederData): Map[String, Any] = data
```
mlaflamm: Should returns `ContextData`

mlaflamm:
+1

PR #150, opened 2013-10-15
A fix that shouldn't be one.

oliviertoupin:
Seems to fix problem with ci1. ArrayBuffer was probably causing more GC.

ListBuffer is the better theorical choice anyway, but ArrayBuffer, proved ok when doing the perf test. I guess I was wrong.


oliviertoupin:
@manu @alex-wajam

mlaflamm:
+1

coderunner:
Shouldn't we revert that. I don't see why ListBuffer is a better choice since we make a get-by-index.

PR #151, opened 2013-10-15
Won't fix the build but nice catch, so I want to commit anyway

oliviertoupin:
@mlaflamm @alex-wajam


alexbergeron:
+1 (I prefer the .asScala way too)

oliviertoupin:
It's not that I prefer that way, it's that it will prevent a shit ton of MapWrapper

alexbergeron:
I said it mostly as a bonus.

It's more explicit and just as readable. 

I know I use lots of implicits theses days, but explicit will always be better than implicit

alexbergeron:
(Second line of Zen of Python)

PR #152, opened 2013-10-15
Ignore the benchmark for now since it's sucking too much CPU.

oliviertoupin:


alexbergeron:
+1

coderunner:
+1, we will then find this test a proper home.

PR #153, opened 2013-10-15
Reverted to ArrayBuffer and using sizeHint for decoding

oliviertoupin:


oliviertoupin:
@alex-wajam

alexbergeron:
Did you mean @coderunner ???

oliviertoupin:
If you could use the same username as on IRC I would be SO happy.

alexbergeron:
+1

PR #154, opened 2013-10-15
Using ScalaTest tags to filter SlowTest and run them differently

oliviertoupin:


oliviertoupin:
@coderunner @alex-wajam @mlaflamm  @clementgarnier

oliviertoupin:
http://www.scalatest.org/user_guide/using_the_runner#filtering
http://doc.scalatest.org/1.9.1/org/scalatest/FunSpec.html @ "Tagging tests"

oliviertoupin:
http://code.google.com/p/scalatest/source/browse/tags/release-1.9.1-for-scala-2.10/src/main/scala/org/scalatest/Tag.scala

oliviertoupin:
Running the normal tests -> sbt test
Running the slow tests -> sbt slow:test

mry-core/src/test/scala/com/wajam/mry/package.scala
```diff
@@ -0,0 +1,7 @@
+package com.wajam
+
+import org.scalatest.Tag
+
+package object mry {
+  object SlowTest extends Tag("SlowTest")
```
alexbergeron, line 6: Why did you put in inside package.scala?

oliviertoupin:
An improvement would be to use http://code.google.com/p/scalatest/source/browse/#svn%2Ftrunk%2Fsrc%2Fmain%2Fjava%2Forg%2Fscalatest%2Ftags

But it's not in any release, or have something like it in ScalaCommons-Test (Because it requires scalatests we don't want to put it in the real thing)

alexbergeron:
Looks good, but I'd like to let somebody else (maybe @coderunner) to look at it one last time before merging

mry-core/src/test/scala/com/wajam/mry/package.scala
oliviertoupin, line 6: I guess I could have put it in test/Tags.scala... But we should anyway but it in some kinds of commons, along with CPUTest, NetworkTest, DiskTest, since those are not specific.

alexbergeron, line 6: I don't think the global package.scala file is a good common for theses. They could still end up in the same file (I like Tags.scala).

And they're not in any commons right now, but might always be extracted one day. commons-tests maybe?

coderunner:
This complexifies the build a lot and add custom configs just to run a single test. Why not just put the test along with future performance tests in Integration tests instead?

coderunner:
Just to be clear, I'm not against tagging the test. It is more the build modifications that I think are not worth it especially because we can solve the problem in a much simpler way.

If we decide to use a Tag, then I would move the tag definition to scala-commons (commons-test) in com.wajam.commons.tags and call it PerformanceTest instead of SlowTest.

oliviertoupin:
@coderunner I agree but:

A - I wanted to prove and show it's possible
B - I see it as a first step towards integrating tags in tests.
C - I have a longer term plan that I wish to talk to you about when you have 5 minutes.

@everyone-else

Any other suggestions, concerns, comments?

PR #155, opened 2013-10-15
Apply limit in outer clause of getMultiple, but avoid full table scan

alexbergeron:
@coderunner @mlaflamm @appaquet

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
@@ -175,8 +175,8 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
     val fullTableName = table.depthName("_")
     val outerProjKeys = (for (i <- 1 to table.depth) yield "o.k%1$d".format(i)).mkString(",")
     val innerProjKeys = (for (i <- 1 to table.depth) yield "i.k%1$d".format(i)).mkString(",")
-    val outerWhereKeys = (for (i <- 1 to table.depth) yield "i.k%1$d = o.k%1$d".format(i)).mkString(" AND ")
-    val innerWhereKeys = (for (i <- 1 to accessPath.parts.length) yield "i.k%1$d = o.k%1$d AND i.k%1$d = ?".format(i)).mkString(" AND ")
+    val joinKeys = (for (i <- 1 to table.depth) yield "i.k%1$d = o.k%1$d".format(i)).mkString(" AND ")
+    val innerWhereKeys = (for (i <- 1 to accessPath.parts.length) yield "i.k%1$d = ?".format(i)).mkString(" AND ")
 
     val innerLimitSql = (optOffset, optCount) match {
       case (Some(offset), Some(count)) => f" LIMIT $offset%d, $count%d"
@@ -192,19 +192,19 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
 
     /* Generated SQL looks like:
      *
-     *   SELECT n.max_ts, n.tk, a.ec,a.d, a.k1,a.k2
-     *   FROM `table1_table1_1_data` AS a, (
+     *   SELECT i.max_ts, i.tk, o.ec,o.d, o.k1,o.k2
+     *   FROM `table1_table1_1_data` AS o, (
      *     SELECT i.tk, MAX(i.ts) AS max_ts, i.k1,i.k2
      *     FROM `table1_table1_1_index` AS i, `table1_table1_1_data` AS o
-     *     WHERE i.tk = ? AND i.k1 = ? AND i.ts <= ?
-     *     AND o.tk = i.tk AND i.k1 = o.k1 AND i.k2 = o.k2 AND i.ts = o.ts
-     *     AND o.d IS NOT NULL
+     *     WHERE i.tk = ? AND i.k1 = o.k1 AND i.k1 = ? AND i.ts <= ?
+     *     AND i.ts = o.ts AND o.tk = i.tk
      *     GROUP BY i.tk, i.k1,i.k2
+     *     HAVING MAX(i.ts) = MAX(IF(o.d IS NOT NULL,i.ts,0))
      *     LIMIT ?
-     *   ) AS n
-     *   WHERE a.tk = n.tk AND n.k1 = a.k1 AND n.k2 = a.k2
-     *   AND a.ts = n.max_ts
-     *   AND a.d IS NOT NULL;
+     *   ) AS i
+     *   WHERE o.tk = i.tk AND i.k1 = o.k1 AND i.k2 = o.k2
+     *   AND o.ts = i.max_ts
+     *   AND o.d IS NOT NULL;
      */
     val sql = """
         SELECT i.max_ts, i.tk, o.ec, o.d, %1$s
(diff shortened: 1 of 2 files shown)


mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
```diff
@@ -176,13 +179,22 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
...
+    val innerLimitSql = optCount match {
+      case Some(count) => f"LIMIT 0, ${(optOffset.map(_.toInt).getOrElse(0) + count.toInt) * innerLimitMagnitude}%d"
+      case _ => ""
+    }
+
+    val outerLimitSql = (optOffset, optCount) match {
+      case (Some(offset), Some(count)) => f" LIMIT $offset%d, $count%d"
+      case (Some(offset), None) =>
+        // MySQL doesn't support limit on offset without a count
+        f" LIMIT $offset%d, $mysqlTheoricalMaxNbOfRows%s"
```
alexbergeron, line 185: > With one argument, the value specifies the number of rows to return from the beginning of the result set
> To retrieve all rows from a certain offset up to the end of the result set, you can use some large number for the second parameter. This statement retrieves all rows from the 96th row to the last:
> 
> ``` sql
> SELECT * FROM tbl LIMIT 95,18446744073709551615;
> ```

http://dev.mysql.com/doc/refman/5.0/en/select.html

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
```diff
@@ -169,6 +169,9 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
                   optCount: Option[Long] = None): RecordIterator = {
     assert(accessPath.length >= 1)
 
+    val innerLimitMagnitude = 100
+    val mysqlTheoricalMaxNbOfRows = "18446744073709551615"
```
alexbergeron, line 172: Taken from http://stackoverflow.com/questions/2716232/maximum-number-of-records-in-a-mysql-database-table/8586081

The answer is bad, but I'm sure we'll never get there before sharding

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
```diff
@@ -209,11 +222,11 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
...
         WHERE o.tk = i.tk AND %6$s
         AND o.ts = i.max_ts
-              """.format(outerProjKeys, fullTableName, innerProjKeys, innerWhereKeys, limitSql, outerWhereKeys)
+        %7$s
+        %8$s
+              """.format(outerProjKeys, fullTableName, innerProjKeys, innerWhereKeys, innerLimitSql, outerWhereKeys, excludeDeleted, outerLimitSql)
 
 
-    if (!includeDeleted)
-      sql += " AND o.d IS NOT NULL "
```
alexbergeron, line 216: Changed this because the string was not built in a consistent way IMHO.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
appaquet, line 172: Holy cow

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
```diff
@@ -176,13 +179,22 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
     val outerWhereKeys = (for (i <- 1 to table.depth) yield "i.k%1$d = o.k%1$d".format(i)).mkString(" AND ")
     val innerWhereKeys = (for (i <- 1 to accessPath.parts.length) yield "i.k%d = ?".format(i)).mkString(" AND ")
 
-    val limitSql = (optOffset, optCount) match {
-      case (Some(offset), Some(count)) => "LIMIT %d, %d".format(offset, count)
-      case (Some(offset), None) => "LIMIT %d".format(offset)
-      case (None, Some(count)) => "LIMIT 0, %d".format(count)
+    val innerLimitSql = optCount match {
+      case Some(count) => f"LIMIT 0, ${(optOffset.map(_.toInt).getOrElse(0) + count.toInt) * innerLimitMagnitude}%d"
```
appaquet: That makes hell of a string, don't you think? For lisbility, I'd rather get the formula to another variable and us a normal string interpolation.

appaquet:
I though we had agreed on the #2 in my suggestions through email? This is the #1 and may not yield right results if number of deleted is 100x number of rows in the inner, no? Plus, I was getting an order magnitude slower with a x1000 on the internal query in my bench, so we may have to test them first...

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
alexbergeron: Yeah, I was trying to find a way, but this string interpolation is way too long, should at leat extract part of it to a val.

alexbergeron:
Updated

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
```diff
@@ -169,51 +169,60 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
...
-     *   AND i.k1 = o.k1
-     *   AND o.ts = i.max_ts
-     *   AND o.d IS NOT NULL
+     *   SELECT n.max_ts, n.tk, a.ec,a.d, a.k1,a.k2
+     *   FROM `table1_table1_1_data` AS a, (
+     *     SELECT i.tk, MAX(i.ts) AS max_ts, i.k1,i.k2
+     *     FROM `table1_table1_1_index` AS i, `table1_table1_1_data` AS o
+     *     WHERE i.tk = ? AND i.k1 = ? AND i.ts <= ?
+     *     AND o.tk = i.tk AND i.k1 = o.k1 AND i.k2 = o.k2 AND i.ts = o.ts
+     *     AND o.d IS NOT NULL
```
appaquet: Misleading comment ;) The IS NOT NULL isn't in the inner select anymore. You should replace by the HAVING instead

alexbergeron: Was trying to push too many PR before leaving :wink:

mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala
```diff
@@ -169,51 +169,60 @@ class MysqlTransaction(private val storage: MysqlStorage, private val context: O
...
+     *     GROUP BY i.tk, i.k1,i.k2
+     *     LIMIT ?
+     *   ) AS n
+     *   WHERE a.tk = n.tk AND n.k1 = a.k1 AND n.k2 = a.k2
+     *   AND a.ts = n.max_ts
+     *   AND a.d IS NOT NULL;
      */
-    var sql = """
```
appaquet, line 201: Woohoo!!

appaquet:
+1 after comment

alexbergeron:
@appaquet - Looking at the code again, I don't think I'm doing the join on the last key. looking at the line

``` scala
val innerWhereKeys = (for (i <- 1 to accessPath.parts.length) yield "i.k%1$d = o.k%1$d AND i.k%1$d = ?".format(i)).mkString(" AND ")
```

would generate, for a 2-level table, something like 

``` sql
i.k1 = o.k1 AND i.k1 = ?
```

while it should be

``` sql
i.k1 = o.k1 AND i.k2 = o.k2 AND i.k1 = ?
```

appaquet:
Yep!

+1

PR #156, opened 2013-11-19
Log MRY transactions errors at MRY level, avoid propagating to NRV

alexbergeron:
@coderunner @mmimeault

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/Database.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/Database.scala
@@ -4,7 +4,7 @@ import api.MryCodec
 import execution._
 import storage.Storage
 import com.wajam.commons.Logging
-import com.wajam.nrv.TimeoutException
+import com.wajam.nrv.{RemoteException, TimeoutException}
 import com.yammer.metrics.scala.Instrumented
 import com.wajam.nrv.service._
 import com.wajam.tracing.Traced
@@ -142,18 +142,17 @@ class Database(serviceName: String = "database")
       context.commit()
       transaction.reset()
 
+      req.reply(
+        null,
+        data = values
+      )
     } catch {
       case e: Exception => {
         debug("Got an exception executing transaction", e)
         context.rollback()
-        throw e
+        req.replyWithError(new RemoteException("Exception executing transaction", e))
       }
     }
-
-    req.reply(
-      null,
-      data = values
-    )
   }
 }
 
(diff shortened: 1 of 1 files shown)


mry-core/src/main/scala/com/wajam/mry/Database.scala
```diff
@@ -142,18 +142,17 @@ class Database(serviceName: String = "database")
...
+      req.reply(
+        null,
+        data = values
+      )
     } catch {
       case e: Exception => {
         debug("Got an exception executing transaction", e)
         context.rollback()
-        throw e
+        req.replyWithError(new RemoteException("Exception executing transaction", e))
```
mmimeault: Could't just be thrown(replied) without be wrapped in a Remote exception?

alexbergeron: I wrapped it inside a RemoteException to replicate current behavior (as seen in NRV)

alexbergeron: I like having it wrapped inside a RemoteException, it makes it clear that the exception occured on another server and not locally on the client side.

But I'd prefer hearing @appaquet opinion on it on a design/architectural perspective.

coderunner: I think it should not be wrapped in a RemoteException otherwise replyWitError should be the one wrapping it. 

But if we make that change, we need to check code that was explicitly matching RemoteException when calling MRY code and adjust it correctly, but I doubt there will be many instances.

alexbergeron: From what I've seen, there is only one instance right now.

alexbergeron:
Updated

appaquet:
Sorry for the late answer. Maybe you could explain why we don't want the errors to propagate to NRV?

As Felix said, it should be the original exception and then NRV handles the transformation to RemoteException. In this case, we should use the original exception and NRV should make sure it's wrapped into a RemoteException.

coderunner:
@appaquet In this case, I think we want MRY to officially reply with the exception and not let NRV catch-all mechanism to wrap it in a generic RemoteException. This way, the caller will get the original exception in the callback instead of having to retreive the original from the RemoteException.

NRV code should stay there to ensure that any exception that slips through an Action execution results in a reply to the caller.

appaquet:
I agree the client must throw his own exception as much as possible, but the problem is that not all exceptions are serializable. It's NRV user's responsibility to make sure their exceptions can be wired. The idea behind the remote exception was that it could be serializable no matter which exception would be wrapped (this feature isn't implemented though). So, I agree we should be using specific exceptions as much as possible and use them in our implementation if they need to be managed differently in the client's code. So, I backtrack my comment. It does make sense now ;)

coderunner:
+1

PR #157, opened 2014-01-29
Update ScalaTest to 2.0

alexbergeron:


mry-core/src/it/scala/com/wajam/mry/storage/mysql/FeederTestHelper.scala
```diff
@@ -3,6 +3,7 @@ package com.wajam.mry.storage.mysql
 import com.wajam.spnl.feeder.Feeder
 import com.wajam.commons.Closable
 import com.wajam.spnl.feeder.Feeder.FeederData
+import scala.language.implicitConversions
```
mmimeault, line 6: why do we need that now ?

alexbergeron, line 6: Compiler warning

mry-core/src/test/scala/com/wajam/mry/api/protobuf/TestMryCodec.scala
```diff
@@ -21,7 +21,7 @@ class TestMryCodec extends FunSuite with BeforeAndAfter with ShouldMatchers {
     val bytes = codec.encode(n)
     val n2 = codec.decode(bytes)
 
-    n2 should equal(null)
+    n2 should === (null)
```
mmimeault, line 24: equal not supported anymore ??
why (null)... could be `n2 should === null` ?

alexbergeron, line 24: equal is still supported, but this didn't compiled for some reason.

I kept the parens as suggested by http://www.scalatest.org/user_guide/using_matchers#thosePeskyParens

> Although you don't always need them, you can choose to always put parentheses around right-hand values

mmimeault:
+1

PR #158, opened 2014-02-13
Partial update support in MRY

alexbergeron:
@jgagnon1 @mlaflamm


alexbergeron:
After some thinking and some discussion with @mlaflamm (that happened before sending this PR), I have some design questions. @appaquet could also answer theses (by this, I mean, @appaquet, you are our only hope)
- Should a set on a `RecordValue` returns a new `RecordValue` wrapping the updated `MapValue`? Personally, I would prefer if we fetched the last value inserted in MySQL (in the same transaction), to be sure to always get the latest record (through, if all ofm this works, we won't have to poke MySQL afterwards) and for API consistency.
- Should set on a `RecordValue` allows to take only one parameter, of type `MapValue`, that would allow to update multiple fields without going back and forth to MySQL? I strongly think so, to the point where I think this should be the only accepted signature. (Me and manu did not had time to discuss this)

alexbergeron:
Will need to implement set on MapValue that returns a new instance of MapValue without calling MySQL.

PR #159, opened 2014-03-11
MRY record multilevel caching

mlaflamm:

PR #160, opened 2014-03-12
Add cache metrics

mlaflamm:
Also include a couple of critical bug fixes:
- Using record empty path in storage cache
- Support resurrected record (maybe not that critical since we probably never do it in prod)

<b>TO BE MERGED INTO #159</b>

### DIFF
--- a/mry-core/src/main/scala/com/wajam/mry/storage/mysql/cache/CacheMetrics.scala
+++ b/mry-core/src/main/scala/com/wajam/mry/storage/mysql/cache/CacheMetrics.scala
@@ -1,0 +1,67 @@
+package com.wajam.mry.storage.mysql.cache
+
+import com.yammer.metrics.scala.Instrumented
+import com.wajam.mry.storage.mysql.Table
+
+trait CacheMetrics extends Instrumented {
+  lazy val hitsMeter = new TableMetrics(descendants = true, scope => metrics.meter("hits", "hits", scope))
+  lazy val missesMeter = new TableMetrics(descendants = true, scope => metrics.meter("misses", "misses", scope))
+
+  lazy val evictionExpiredCounter = new TableMetrics(descendants = false, scope => metrics.counter("eviction-expired", scope))
+  lazy val evictionSizeCounter = new TableMetrics(descendants = false, scope => metrics.counter("eviction-size", scope))
+
+  val cacheCurrentSizeGauge = new TableGauges[Int]("cache-current-size")
+  val cacheMaxSizeGauge = new TableGauges[Int]("cache-max-size")
+
+  protected def resetMetrics(): Unit = {
+    cacheCurrentSizeGauge.reset()
+    cacheMaxSizeGauge.reset()
+  }
+
+  class TableMetrics[T](descendants: Boolean, newMetric: (String) => T) {
+    private var cacheMetrics: Map[String, T] = Map()
+
+    def apply(table: Table): Seq[T] = {
+      val seq = getOrAddMetric(CacheMetrics.TotalScope) :: getOrAddMetric(table.getTopLevelTable.name) :: Nil
+      if (descendants) getOrAddMetric(table.toString) :: seq else seq
+    }
+
+    private def getOrAddMetric(scope: String): T = {
+      cacheMetrics.get(scope) match {
+        case Some(metric) => metric
+        case None => {
+          val metric = newMetric(scope)
+          cacheMetrics += scope -> metric
+          metric
+        }
+      }
+    }
+  }
+
+  class TableGauges[T: Numeric](name: String) {
+    private var cacheGauges: Map[String, Function0[T]] = Map()
+
+    def addTable(table: Table, value: => T): Unit = {
+      addGauge(table.getTopLevelTable.name, value)
+      metrics.gauge(name, CacheMetrics.TotalScope) { cacheGauges.values.map(_()).sum }
+    }
+
+    def reset() = {
+      removeGauge(CacheMetrics.TotalScope)
+      cacheGauges.keys.foreach(removeGauge)
+      cacheGauges = Map()
+    }
+
+    private def addGauge(scope: String, value: => T): Unit = {
+      metrics.gauge(name, scope) { value }
+      cacheGauges += scope -> (() => value)
+    }
+
+    private def removeGauge(scope: String) = metricsRegistry.removeMetric(metrics.klass, name, scope)
+  }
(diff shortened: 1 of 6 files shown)


mry-core/src/main/scala/com/wajam/mry/storage/mysql/cache/HierarchicalCache.scala
```diff
@@ -29,14 +35,17 @@ class HierarchicalTableCache(expireMs: Long, maximumSize: Int) extends TableCach
...
     Option(cache.getIfPresent(path))
   }
 
   def put(path: AccessPath, record: Record) = {
-    trace(s"put(): $record")
-    keys.add(record.accessPath)
-    cache.put(record.accessPath, record)
+    trace(s"put(): $path, $record")
+    keys.add(path)
+    cache.put(path, record)
```
mlaflamm, line 48: MRY does not fill `record.accessPath`. The original code was lousy anyway.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/cache/CacheMetrics.scala
```diff
@@ -0,0 +1,67 @@
...
+
+    private def getOrAddMetric(scope: String): T = {
+      cacheMetrics.get(scope) match {
+        case Some(metric) => metric
+        case None => {
+          val metric = newMetric(scope)
+          cacheMetrics += scope -> metric
+          metric
+        }
+      }
```
alexbergeron, line 34:
``` scala
cacheMetrics.get(scope) getOrElse {
  val metric = newMetric(scope)
  cacheMetrics += scope -> metric
  metric
}
```

mry-core/src/main/scala/com/wajam/mry/storage/mysql/cache/CacheMetrics.scala
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  class TableGauges[T: Numeric](name: String) {
+    private var cacheGauges: Map[String, Function0[T]] = Map()
```
alexbergeron: `() => T`

mry-core/src/main/scala/com/wajam/mry/storage/mysql/cache/HierarchicalCache.scala
alexbergeron, line 48: Should create an issue...

mry-core/src/main/scala/com/wajam/mry/storage/mysql/cache/HierarchicalCache.scala
```diff
@@ -52,18 +61,26 @@ class HierarchicalTableCache(expireMs: Long, maximumSize: Int) extends TableCach
...
 
   private object CacheRemovalListener extends RemovalListener[AccessPath, Record] {
-    def onRemoval(notification: RemovalNotification[AccessPath, Record]) {
+    def onRemoval(notification: RemovalNotification[AccessPath, Record]): Unit = {
       val cause = notification.getCause
-      trace(s" onRemoval(): ${notification.getCause}, evicted=${wasEvicted(cause)}, $notification")
-      if (wasEvicted(cause)) {
-        keys.remove(notification.getKey)
+      trace(s" onRemoval(): $cause, evicted=${wasEvicted(cause)}, $notification")
+      cause match {
```
alexbergeron, line 74: Personally, I would do

``` scala
notification.getCause match {
```

But it's still a question of personal preference

mry-core/src/main/scala/com/wajam/mry/storage/mysql/cache/TransactionCache.scala
```diff
@@ -41,8 +45,15 @@ class TransactionCache(getTableCache: (Table) => TableCache[Record]) extends Log
 
     trxTableCaches.foreach {
       case (table, trxTableCache) => trxTableCache.toIterable.foreach {
-        case (path, CachedValue(Some(rec), _)) => getTableCache(table).put(path, rec)
-        case (path, CachedValue(None, _)) => getTableCache(table).invalidate(path)
+        case (path, CachedValue(Some(rec), _, invalidateDescendants)) if invalidateDescendants => {
```
alexbergeron, line 48:
``` scala
case (path, CachedValue(Some(rec), _, true)) => {
```

(But I'm not 100% it works)

mry-core/src/main/scala/com/wajam/mry/storage/mysql/cache/HierarchicalCache.scala
mlaflamm, line 48: This kind of saves memory but I agree with you and I will do it.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/cache/HierarchicalCache.scala
mlaflamm, line 74: I would do it if it wasn't also used twice in the `trace`

mry-core/src/main/scala/com/wajam/mry/storage/mysql/cache/HierarchicalCache.scala
alexbergeron, line 48: It could always be a def

mry-core/src/main/scala/com/wajam/mry/storage/mysql/cache/TransactionCache.scala
mlaflamm, line 48: I confirm it works and your version is prettier but I think the original code is clearer because of the `invalidateDescendants` name. Unless you have a strong opinion on this, I'll leave it as is.

alexbergeron, line 48: I think you're right, booleans are not very clear in general...

alexbergeron:
+1

mry-core/src/main/scala/com/wajam/mry/storage/mysql/cache/HierarchicalCache.scala
mlaflamm, line 48: OMG, the [Record](https://github.com/wajam/mry/blob/master/mry-core/src/main/scala/com/wajam/mry/storage/mysql/MysqlTransaction.scala#L883-L885) class is a scary can of :worm:...

PR #163, opened 2014-03-24
Fix dry mode when using partial updates

alexbergeron:
@mlaflamm @jgagnon1 @appaquet

### DIFF
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/MysqlStorageFixture.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/MysqlStorageFixture.scala
@@ -26,6 +26,16 @@ trait MysqlStorageFixture {
           context.rollback()
       }
     }
+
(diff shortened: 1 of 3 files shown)


appaquet:
Pretty sure we should make more tests with dryMode. There may be other bugs laying around.

+1 from me for the changes though

mlaflamm:
I agree with @appaquet, we can probably make `MysqlStorageFixture.exec()` run once in dry mode followed by the standard mode for all tests. 

+1 for this pull request. I added the story BE-446 for adding extra <b>dry mode</b> tests.

alexbergeron:
I agree that we're missing more integration tests with dry mode, but I have two points I need to make before merging this:
1. I'm not running the actual test that found this bug in dry mode. I should fix this before merging
2. There could be a simpler fix for this issue, possibly in `TableValue`

alexbergeron:
Updated

mlaflamm:
+1

PR #164, opened 2014-04-15
RunOnce Feeder. Should be persisted and controllable.

alexbergeron:
@mlaflamm @jgagnon1

### DIFF
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/FeederTestHelper.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/FeederTestHelper.scala
@@ -4,6 +4,7 @@ import com.wajam.spnl.feeder.Feeder
 import com.wajam.commons.Closable
 import com.wajam.spnl.feeder.Feeder.FeederData
 import scala.language.implicitConversions
+import com.wajam.nrv.service.{TokenRange, TokenRangeSeq}
 
 object FeederTestHelper {
   implicit def feederToIterator(feeder: Feeder): Iterator[Option[FeederData]] with Closable = {
@@ -17,14 +18,33 @@ object FeederTestHelper {
     }
   }
 
-  implicit def onceFeederToIterator(feeder: Feeder with TableOnceFeeder): Iterator[Option[FeederData]] with Closable = {
+  /**
+   * Test feeder which use token as record. When loading records, it simply enumerates all the tokens in the
+   * specified token range.
+   */
+  abstract class TokenFeeder(val tokenRanges: TokenRangeSeq, limit: Int)
+    extends ResumableRecordDataFeeder {
 
-    new Iterator[Option[FeederData]] with Closable {
-      def hasNext = feeder.hasNext
+    var firstRangeLoadCount = 0
 
-      def next() = feeder.next()
+    type DataRecord = Long
 
-      def close() = feeder.kill()
+    val name = "ContinuousTokenFeeder"
+
+    def token(record: Long) = record
+
+    def loadRecords(range: TokenRange, startAfterRecord: Option[Long]) = {
+      firstRangeLoadCount += (if (range == tokenRanges.head && startAfterRecord.isEmpty) 1 else 0)
+
+      (startAfterRecord match {
+        case Some(start) => (start + 1).to(range.end)
+        case None => range.start.to(range.end)
+      }).take(limit)
     }
+
+    def toRecord(data: FeederData) = data.get("token").map(_.toString.toLong)
+
+    def fromRecord(record: Long) = Map("token" -> record)
   }
+
 }
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/FeederTestHelper.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/FeederTestHelper.scala
@@ -29,7 +29,7 @@ object FeederTestHelper {
 
     type DataRecord = Long
 
-    val name = "ContinuousTokenFeeder"
+    val name = "TokenFeeder"
 
     def token(record: Long) = record
 
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/FeederTestHelper.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/FeederTestHelper.scala
@@ -16,4 +16,15 @@ object FeederTestHelper {
       def close() = feeder.kill()
     }
   }
+
+  implicit def onceFeederToIterator(feeder: Feeder with TableOnceFeeder): Iterator[Option[FeederData]] with Closable = {
+
+    new Iterator[Option[FeederData]] with Closable {
+      def hasNext = feeder.hasNext
+
+      def next() = feeder.next()
+
+      def close() = feeder.kill()
+    }
+  }
 }
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/FeederTestHelper.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/FeederTestHelper.scala
@@ -44,7 +44,7 @@ object FeederTestHelper {
 
     def toRecord(data: FeederData) = data.get("token").map(_.toString.toLong)
 
-    def fromRecord(record: Long) = Map("token" -> record)
+    def fromRecord(record: DataRecord): FeederData = Map("token" -> record)
   }
 
 }
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableOnceFeeder.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableOnceFeeder.scala
@@ -17,9 +17,7 @@ class TestTableOnceFeeder extends FunSuite {
    * specified token range.
    */
   class OnceTokenFeeder(tokenRanges: TokenRangeSeq, limit: Int)
-    extends TokenFeeder(tokenRanges, limit) with TableOnceFeeder {
-
-  }
+    extends TokenFeeder(tokenRanges, limit) with TableOnceFeeder
 
   test("should resume from context position") {
     val ranges = Seq(TokenRange(2, 5), TokenRange(10, 12))
@@ -116,8 +114,11 @@ class TestTableOnceFeeder extends FunSuite {
     val records = feeder.take(50).flatten.toList
     records.flatMap(feeder.toRecord).take(15) should be(List(2, 3, 4, 5, 10, 11, 12))
 
+    feeder.isStarted should be(false) // No more values, but not "resetted"
     feeder.stop() should be(true)
+    feeder.isStarted should be(false)
     feeder.stop() should be(false)
+    feeder.isStarted should be(false)
     Thread.sleep(1000)
     feeder.start("") should be(true)
     feeder.start("") should be(false)
@@ -132,6 +133,7 @@ class TestTableOnceFeeder extends FunSuite {
     feeder.start("") should be(true)
     val records = feeder.take(11).flatten.toList
     records.flatMap(feeder.toRecord).take(15) should be(List(2, 3, 4, 5, 10))
+    feeder.isStarted should be(true)
     feeder.stop() should be(true)
     val records2 = feeder.take(5).flatten.toList
     records2.flatMap(feeder.toRecord).take(15) should be(List())
@@ -143,6 +145,7 @@ class TestTableOnceFeeder extends FunSuite {
     feeder.start("") should be(true)
     val records = feeder.take(50).flatten.toList
     records.flatMap(feeder.toRecord).take(15) should be(List())
+    feeder.isStarted should be(true)
     feeder.stop() should be(true)
     val records2 = feeder.take(50).flatten.toList
     records2.flatMap(feeder.toRecord).take(15) should be(List())
--- a/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableOnceFeeder.scala
+++ b/mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableOnceFeeder.scala
@@ -137,4 +137,14 @@ class TestTableOnceFeeder extends FunSuite {
     records2.flatMap(feeder.toRecord).take(15) should be(List())
   }
 
+  test("should support empty feeders") {
+    val feeder = new OnceTokenFeeder(Nil, limit = 1)
+    feeder.init(TaskContext())
+    feeder.start("") should be(true)
+    val records = feeder.take(50).flatten.toList
+    records.flatMap(feeder.toRecord).take(15) should be(List())
+    feeder.stop() should be(true)
+    val records2 = feeder.take(50).flatten.toList
+    records2.flatMap(feeder.toRecord).take(15) should be(List())
+  }
 }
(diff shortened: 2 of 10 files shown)


mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableOnceFeeder.scala
```diff
@@ -0,0 +1,150 @@
...
+
+@RunWith(classOf[JUnitRunner])
+class TestTableOnceFeeder extends FunSuite {
+
+  /**
+   * Test feeder which use token as record. When loading records, it simply enumerates all the tokens in the
+   * specified token range.
+   */
+  class OnceTokenFeeder(val tokenRanges: TokenRangeSeq, limit: Int)
+    extends ResumableRecordDataFeeder with TableOnceFeeder {
```
mlaflamm: The implementation is very, very, very similar to `ContinuousTokenFeeder` defined in `TestTableContinuousFeeder`, only the traits mixed-in are different. I suggest to extract a common abstract class...

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableOnceFeeder.scala
```diff
@@ -0,0 +1,66 @@
...
+    if (hasNext) false else {
+      context.data = context.data.updated("start-date", DateTime.now)
+      true
+    }
+  }
+
+  def reset(): Boolean = {
+    if (hasNext) false else {
+      lastRecord = None
+      currentRange = None
```
mlaflamm, line 48: I doubt this is very thread safe. This method is likely called from a thread different from the SPNL scheduler.

mlaflamm, line 48: This is also true for `start` and `restart`

alexbergeron, line 48: Who needs Thread safety? With all of this mutability, we're probably building a single-threaded application!

...Yeah, we'll have to do better than this. But I don't think making it `synchronized` would do it, since theses values might be modified from the main thread...

alexbergeron, line 48: I've got an idea using a volatile value or an atomic boolean... Can we discuss?

mlaflamm, line 48: Before the addition of these methods, the feeder was always running on the SPNL scheduler thread.

mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableOnceFeeder.scala
alexbergeron: The `Feeder` in `TestTableContinuousFeeder` also has a `firstRangeLoadCount`, but it would just be unused in this case...

alexbergeron:
Updated

mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableOnceFeeder.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  test("feeder should not loop when loaded all data - large load limit") {
+    val ranges = Seq(TokenRange(2, 5), TokenRange(10, 12))
+    val feeder = new OnceTokenFeeder(ranges, limit = 1000)
+    feeder.init(TaskContext())
+    feeder.start()
+
+    val records = feeder.take(50).flatten.toList
+    records.flatMap(feeder.toRecord).take(15) should be(List(2, 3, 4, 5, 10, 11, 12))
+  }
```
jgagnon1, line 94: Not sure to see how you check that feeder is not looping on large limit ?

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableFeederByToken.scala
```diff
@@ -0,0 +1,49 @@
...
+import com.wajam.spnl.TaskContext
+
+/**
+ * Feeder trait which ease creation of feeders who iterate over table records sequentially per range of tokens.
+ * When started it resume from the last acknowledged record saved in the task context.
+ */
+private[mysql] trait TableFeederByToken extends CachedDataFeeder with ResumableRecordDataFeeder with Logging {
+
+  private var currentContext: TaskContext = null
+  private[mysql] var lastRecord: Option[DataRecord] = None
```
jgagnon1, line 15: since the class is already private[mysql] do you need it here to ?

mlaflamm, line 15: Right, protected look like more appropriate in that case as it can only be accessed by subclasses.

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableOnceFeeder.scala
```diff
@@ -0,0 +1,103 @@
...
+
+  private def setDone(): Unit = {
+    updateContext(Map(finishedDate -> dateFormatter.print(DateTime.now)))
+  }
+
+  private def emptyResult =
+    (TokenRange(1, 0), lastRecord) // start is after end
+
+  private def updateContext(values: Map[String, String] = Map(), emptyContextToUpdate: Boolean = false): Map[String, String] = {
+    contextDataToUpdate synchronized {
```
mlaflamm: Using the mutable value we want to protect with a lock as the lock monitor is bad! You need to use a value that can never change during the lifetime of this feeder instance.

mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableOnceFeeder.scala
alexbergeron, line 94: Maybe the name is not really appropriate...

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableFeederByToken.scala
alexbergeron, line 15: I think this might leak it in projects using it, extending a trait that extends this trait...

alexbergeron, line 15: Ideally, I would only allow subclasses in the `mysql` package to access this.

alexbergeron, line 15: After some check, `private[mysql]` does look like the better choice.

alexbergeron:
Updated

jgagnon1:
Much locking! I guess that is because it can be called from CRTX and SPNL right ? However, those will not be called that much I suppose so it's fine. If @mlaflamm suggested that to you, I'll be waiting for him to +1

alexbergeron:
Yeah, basically, I'm locking on anything that might be executed on another thread than the job thread.. I'm trying to reduce locking to a minimum in the job.

mry-core/src/it/scala/com/wajam/mry/storage/mysql/FeederTestHelper.scala
```diff
@@ -16,4 +17,34 @@ object FeederTestHelper {
...
+   * specified token range.
+   */
+  abstract class TokenFeeder(val tokenRanges: TokenRangeSeq, limit: Int)
+    extends ResumableRecordDataFeeder {
+
+    var firstRangeLoadCount = 0
+
+    type DataRecord = Long
+
+    val name = "ContinuousTokenFeeder"
```
mlaflamm: `"TokenFeeder"`?

mry-core/src/main/scala/com/wajam/mry/storage/mysql/Job.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import com.wajam.nrv.utils.Startable
+
+/**
+ * Marker trait allowing the definition of a Runnable Once job.
+ * Should be mixed in to a Feeder and to the SPNL Task running it
+ */
+trait Job {
+  def isStarted: Boolean
+  def getCurrentId: String
```
mlaflamm: `def currentJobId: String`?

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableOnceFeeder.scala
```diff
@@ -0,0 +1,104 @@
...
+        case None => emptyResult
+      }
+    } else emptyResult
+  }
+}
+
+object TableOnceFeeder {
+  private val startDate = "start-date"
+  private val finishedDate = "finished-date"
+  private val jobId = "job-id"
```
mlaflamm: `StartDate`, `FinishedDate` and `JobId`

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableOnceFeeder.scala
```diff
@@ -0,0 +1,104 @@
...
+  def getCurrentId: String = Lock.synchronized(name + getJobId)
+
+  def isStarted: Boolean = hasNext
+
+  abstract override def toContextData(data: Feeder.FeederData): TaskContext.ContextData = super.toContextData(data) ++
+    Map(startDate -> getStartDate, finishedDate -> getFinishedDate, jobId -> getJobId)
+
+  private[mysql] def getLoadPosition: (TokenRange, Option[DataRecord]) = {
+    context.data ++= getContextToUpdate(emptyContextToUpdate = true)
+    if (hasNext) {
```
mlaflamm: It look like the whole method need to synchronized. What if `start` or `stop` are executed between these two lines?

```
  context.data ++= getContextToUpdate(emptyContextToUpdate = true)
  if (hasNext) {
```

alexbergeron: That might allow me to check less things...

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableOnceFeeder.scala
```diff
@@ -0,0 +1,104 @@
...
+    (TokenRange(1, 0), lastRecord) // start is after end
+
+  private def updateContext(values: Map[String, String]): Unit = Lock.synchronized {
+    val updated = contextDataToUpdate ++ values
+    contextDataToUpdate = updated
+    updated
+  }
+
+
+  private def getContextToUpdate(emptyContextToUpdate: Boolean = false): Map[String, String] = Lock.synchronized {
```
mlaflamm: I must admit I don't get the purpose of `emptyContextToUpdate`. This method is invoked twice, once with the flag and once without. Wouldn't it be better to just in-line this method?

mry-core/src/main/scala/com/wajam/mry/storage/mysql/TableOnceFeeder.scala
```diff
@@ -0,0 +1,104 @@
...
+            setDone()
+            None
+          }
+          case None => None
+        }
+      }
+
+      loadRange match {
+        case Some(range) => (range, lastRecord)
+        case None if hasNext && !getContextToUpdate().contains(finishedDate) => (tokenRanges.head, None)
```
mlaflamm: Why `hasNext` is not sufficient and that a check of `finishedDate` is also necessary?

alexbergeron: Because `hasNext` only checks `context.data`, which hasn't been updaterd yet, since `setDone` affects only the temporary `contextDataToUpdate`

mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableOnceFeeder.scala
```diff
@@ -0,0 +1,140 @@
...
+import org.scalatest.junit.JUnitRunner
+import org.scalatest.FunSuite
+import org.scalatest.Matchers._
+import com.wajam.nrv.service.{TokenRangeSeq, TokenRange}
+import com.wajam.spnl.TaskContext
+import org.mockito.Mockito._
+import com.wajam.mry.storage.mysql.FeederTestHelper._
+
+@RunWith(classOf[JUnitRunner])
+class TestTableOnceFeeder extends FunSuite {
```
mlaflamm, line 13: Please add a test with an empty feeder to ensure it complete as expected.

alexbergeron, line 13: Done

mry-core/src/it/scala/com/wajam/mry/storage/mysql/FeederTestHelper.scala
alexbergeron: Forgotten

alexbergeron: You won't like it, but I'll force a push with an updated commit

alexbergeron:
Updated

mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableOnceFeeder.scala
```diff
@@ -0,0 +1,150 @@
...
+    records2.flatMap(feeder.toRecord).take(15) should be(List())
+  }
+
+  test("should support empty feeders") {
+    val feeder = new OnceTokenFeeder(Nil, limit = 1)
+    feeder.init(TaskContext())
+    feeder.start("") should be(true)
+    val records = feeder.take(50).flatten.toList
+    records.flatMap(feeder.toRecord).take(15) should be(List())
+    feeder.stop() should be(true)
```
mlaflamm: I think you should verify `isStarted` before manually stopping it to ensure it stopped naturally beforehand.

mlaflamm:
The `TableOnceFeeder` need to pass the JobId with every Record.

alexbergeron:
> The TableOnceFeeder need to pass the JobId with every Record.

`TableOnceFeeder` can't define the appropriate method (`fromRecord`), even using `abstract override`, since it doesn't know about the actual type of `DataRecord`

alexbergeron:
Disregard what I've just said an hour ago

alexbergeron:
Updated

mry-core/src/it/scala/com/wajam/mry/storage/mysql/TestTableOnceFeeder.scala
```diff
@@ -0,0 +1,153 @@
...
+    records2.flatMap(feeder.toRecord).take(15) should be(List())
+  }
+
+  test("should support empty feeders") {
+    val feeder = new OnceTokenFeeder(Nil, limit = 1)
+    feeder.init(TaskContext())
+    feeder.start("") should be(true)
+    val records = feeder.take(50).flatten.toList
+    records.flatMap(feeder.toRecord).take(15) should be(List())
+    feeder.isStarted should be(true)
```
mlaflamm: Why true if the table is empty? After the table is completely iterated once, `isStarted` should be false. Right? Perhaps empty tables are special. I think it worth verifying `isStarted` in every tests...

mlaflamm: I think that `TokenFeeder` is just not appropriate for this test.

alexbergeron:
> I think that TokenFeeder is just not appropriate for this test.

Explain?

And if it's not appropriate, I waisted a sprint :crying_cat_face:

mlaflamm:
I agree, your new F&O shirt doesn't fit you <b>waist</b>...

alexbergeron:
Updated :crying_cat_face:

mlaflamm:
+1

